PR #24, opened 2021-07-20
Use Poetry in build scripts and CI

bitner:
Remove remnants (setup.py, requirements-dev.txt) and use Poetry for all builds and for pushing to PyPi.
Fixes #21 

Add release trigger for git tags that creates a release from Tags and also posts docker image to ghcr.io 
Fixes #25 
Fixes #19
Fixes #6

PR #25, opened 2021-07-20
CI workflow to push docker image to github packages

### LINKED ISSUE #19: Push to dockerhub during CI publish workflow
As follow up work on #18, we need to add the ability to push docker images to the cipublish script. Pypi publishing is already handled by cipublish, and the addition of publishing docker images will take care of most of the functionality in the `Makefile`, which should be removed once this and any other functionality is covered by `scripts/`

bitner:
Push releases to GitHub packages docker registry.

Fixes #19

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - "*"
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
 jobs:
   release:
     name: release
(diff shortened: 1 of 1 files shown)


lossyrob:
@bitner looking to get a 0.2.8 release out as the ingestion is broken in 0.2.7 and it's blocking testing of pgstac on stac-fastapi; I'm fine manually tweaking version numbers and will have a PR up to review in a bit that should be good for the release; is there any reason not to merge this to also have a 0.2.8 image to work with from gcr?

lossyrob:
> it's blocking testing of pgstac on stac-fastapi

This may not be true, looks like the testing env is generally botched on my machine; but generally would be good to get a release with https://github.com/stac-utils/pgstac/commit/26df853ba27eeef830974e0d6f6d2511c4fda59f out

vincentsarago:
@bitner any reason not to push to dockerhub and GitHub?

bitner:
> @bitner any reason not to push to dockerhub and GitHub?
> 
> ref: https://github.com/developmentseed/titiler/blob/master/.github/workflows/ci.yml#L142-L152

@vincentsarago my immediate response is lazily that ghcr.io doesn't require me to find any additional credentials

bitner:
Closing this PR, I wrapped this into #24

lossyrob:
@vincentsarago also Dockerhub has user account restrictions # for free organization (e.g. stac-utils on dockerhub), and doesn't allow you to publish with org-specific credentials. We got in a situation where we thought we may have to publish to stac-utils dockerhub with a developers account creds, I went to make a bot account instead, but we were already at the limit for the org...then we found out GCR would solve all of these problems and have been focusing on that as a docker publishing platform since.

PR #26, opened 2021-07-20
Fix issue with logic around type of object returned by smart_open

lossyrob:
This fixes an issue introduced in #18 with the usage of `isinstance` to check the return type of `smart_open.open`.
There was a check in place that was made under the assumption the return type would be consistent with what I was seeing in my development environment at time of coding; however in another instance I found this open returned an unexpected type for an
open file.

This change avoids having type checking interfere with runtime logic by assigning the open for file to `Any`, which skips further type checking.

### DIFF
--- a/pypgstac/pypgstac/pypgstac.py
+++ b/pypgstac/pypgstac/pypgstac.py
@@ -1,5 +1,4 @@
 import asyncio
-from io import BufferedIOBase
 import os
 import time
 from typing import Any, AsyncGenerator, Dict, Iterable, Optional, TypeVar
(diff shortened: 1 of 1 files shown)


bitner:
Hey @lossyrob is there a reason related to mypy or something that we can't just use the open within the context? ie
```
with open(file, "rb") as f:
    async with DB(dsn) as conn:
        await load_iterator(f, table, conn, method)
```

lossyrob:
@bitner yeah, smart_open.open describes a return type, but it's `TextIOWrapper | str | None`, which trips up mypy. Opening before the context expression allows type hinting against the return value, and the previous broken logic was around the return type I was seeing from smart_open.open (which was not TextIOWrapper). Seems like an instance where type hinting is used by not actually adhered to, which causes mypy to be cranky. The fix here could be 

```python
with open(file, "rb") as f:  # type:ignore
    async with DB(dsn) as conn:
        await load_iterator(f, table, conn, method)
```

but then the type information around f is not ignored moving forward in the code, and other "type:ignore" tags might be needed. Assigning the return value to `Any` making mypy basically ignore inferring information about the type.

bitner:
uggh. annoying, but makes sense.

bitner (approved):

PR #27, opened 2021-07-22
Migration Tooling / Release Process

bitner:
Creates tooling to create staged database migrations and appropriately set the version number prior to tagging a release.

- Adds stage version script
  - Use poetry to determine version number -- version can be set either explicitly "v0.1.1" or using a relative keyword "patch","minor","major" to determine the next tag
  - Sets version in pyproject.toml
  - Sets version number in pypgstac/__init__.py
  - Generates base migration for the version (all sql required to build db from scratch)
  - Uses Migra to generate a SQL diff between all tagged versions to be used for incremental migrations.
    - Incremental migrations can be destructive, so this file is saved as pgstac.[fromversion]-[toversion].sql.staged. This file must be reviewed and renamed (remove the .staged suffix).
- Adds tests for tagging process
  - Tests that running accumulated incremental migrations gives same result as full sql 
  - Runs PGTap tests both on a database created with the base sql file for the version and an accumulated incrementally built database
  - Makes sure that all incremental migrations created by staging script have been reviewed / had the ".staged" suffix removed

Fixes #28 
Fixes #24 
Fixes #20 
Fixes #7
Fixes #30

PR #29, opened 2021-07-23
Version 0.2.8

lossyrob:
Updates version to 0.2.8. Includes migration scripts from 0.2.7 databases. Note that there were no database changes, but the current migration functionality requires these scripts to be present; this requirement will be fixed in #27 

Also adds a CHANGELOG with changes only from 0.2.7-0.2.8


pypgstac/setup.py
```diff
@@ -6,15 +6,8 @@
...
-              'migrations/pgstac.0.2.4-0.2.7.sql',
-              'migrations/pgstac.0.2.4-0.2.7.sql',
-              'migrations/pgstac.0.2.5-0.2.7.sql',
-              'migrations/pgstac.0.2.5-0.2.7.sql',
-              'migrations/pgstac.0.2.5-0.2.7.sql',
-              'migrations/pgstac.0.2.7.sql',
-              'migrations/pgstac.0.2.7.sql',
-              'migrations/pgstac.0.2.7.sql']}
+ 'pypgstac': ['migrations/pgstac.0.2.5-0.2.8.sql',
+              'migrations/pgstac.0.2.8.sql']}
```
lossyrob, line 10: I believe the setup.py is generated through poetry, but I manually updated this; removed the duplicate entries which didn't seem necessary.

bitner (approved):
Looks good, I'm trying to wrap up #27 which will streamline a lot of this.

PR #32, opened 2021-07-26
add create_items function for bulk insert

### LINKED ISSUE #31: add bulk insert function
In order to add a bulk insert endpoint it would be great to have a `create_items` function 

```sql
CREATE OR REPLACE FUNCTION create_items(datas jsonb) RETURNS VOID AS $$
    INSERT INTO items (content) SELECT * FROM jsonb_array_elements(datas);
    SELECT backfill_partitions();
$$ LANGUAGE SQL SET SEARCH_PATH TO pgstac,public;
```

vincentsarago:
closes #31

### DIFF
--- a/sql/003_items.sql
+++ b/sql/003_items.sql
@@ -79,6 +79,11 @@ CREATE OR REPLACE FUNCTION create_item(data jsonb) RETURNS VOID AS $$
     INSERT INTO items (content) VALUES (data);
 $$ LANGUAGE SQL SET SEARCH_PATH TO pgstac,public;
 
+-- Bulk insert function which takes a sequence of items as input
+CREATE OR REPLACE FUNCTION create_items(datas jsonb) RETURNS VOID AS $$
+    INSERT INTO items (content) SELECT * FROM jsonb_array_elements(datas);
+    SELECT backfill_partitions();
+$$ LANGUAGE SQL SET SEARCH_PATH TO pgstac,public;
 
 CREATE OR REPLACE FUNCTION update_item(data jsonb) RETURNS VOID AS $$
 DECLARE
--- a/test/pgtap.sql
+++ b/test/pgtap.sql
@@ -98,6 +98,7 @@ SELECT is_partitioned('pgstac'::name,'items'::name);
 SELECT has_function('pgstac'::name, 'get_item', ARRAY['text']);
 SELECT has_function('pgstac'::name, 'delete_item', ARRAY['text']);
 SELECT has_function('pgstac'::name, 'create_item', ARRAY['jsonb']);
+SELECT has_function('pgstac'::name, 'create_items', ARRAY['jsonb']);
 SELECT has_function('pgstac'::name, 'update_item', ARRAY['jsonb']);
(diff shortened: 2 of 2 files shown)


sql/003_items.sql
```diff
@@ -79,6 +79,11 @@ CREATE OR REPLACE FUNCTION create_item(data jsonb) RETURNS VOID AS $$
     INSERT INTO items (content) VALUES (data);
 $$ LANGUAGE SQL SET SEARCH_PATH TO pgstac,public;
 
+-- Bulk insert function which takes a sequence of items as input
+CREATE OR REPLACE FUNCTION create_items(datas jsonb) RETURNS VOID AS $$
+    INSERT INTO items (content) SELECT * FROM jsonb_array_elements(datas);
+    SELECT backfill_partitions();
```
vincentsarago, line 85: @bitner is this the best way to update partitions?

test/pgtap.sql
```diff
@@ -98,6 +98,7 @@ SELECT is_partitioned('pgstac'::name,'items'::name);
 SELECT has_function('pgstac'::name, 'get_item', ARRAY['text']);
 SELECT has_function('pgstac'::name, 'delete_item', ARRAY['text']);
 SELECT has_function('pgstac'::name, 'create_item', ARRAY['jsonb']);
+SELECT has_function('pgstac'::name, 'create_items', ARRAY['jsonb']);
 SELECT has_function('pgstac'::name, 'update_item', ARRAY['jsonb']);
 SELECT has_function('pgstac'::name, 'upsert_item', ARRAY['jsonb']);
```
kylebarron, line 103: Maybe we should also have an `upsert_items`? 🤷‍♂️

sql/003_items.sql
bitner: Yeah, that will work just fine. I'm assuming you modeled this off the pypgstac loader code - for the insert case where we fail if we hit a duplicate id, in the case that the partitions are likely to already exist, it's faster to do like this and add the items first (any items that do not have a partition yet will go into the items_default partition) and then backfill. When the partitions don't exist yet for any data, it is going to be slower in this case than following the code in copy_ignore_duplicates and copy_upsert in pypgstac load where we bring the data into a temporary table first, pre-generate any missing partitions (getting the range from the temp table) and then load the data into the items table either using "on conflict do nothing" to ignore dupes or "on conflict do update..." to upsert.

test/pgtap.sql
bitner: @kylebarron there is also the third option which in the pypgstac loader I have as "ignore_duplicates" where it will not overwrite any data, but it will not fail if the id already exists.

vincentsarago:
@bitner CI fails but it's likely a miss configuration because the PR comes from my personal repo. Let me know if there is more to do for it to get merged 🙏

sql/003_items.sql
lossyrob: pg_partman was dropped, so we should remove this line right?

bitner:
Closing this PR as create_items (and update_items) are added as part of PR #39

PR #50, opened 2021-08-25
Ensure connection is closed before exiting

duckontheweb:
If no migrations were necessary, the `run_migration` function had been exiting without closing the open `asyncpg.Connection`. This fixes that issue and wraps the entire function after opening the connection in a try/finally statement to ensure the connection gets closed.


bitner (approved):
Nice catch. Thanks!

PR #52, opened 2021-09-06
add tests for point search

vincentsarago:
This PR adds tests to demonstrate a `bug` with the `geometrysearch` function when passing a geometry without `area` (e.g point). If `exitwhenfull` / `skipcovered` are set to `true` the geometrysearch function will always return `[]` 

ref: https://github.com/stac-utils/titiler-pgstac/blob/master/tests/test_titiler_pgstac.py#L46-L53


test/pgtap/006_tilesearch.sql
```diff
@@ -42,3 +42,19 @@ SELECT results_eq($$
...
     'Test xyzsearch to return feature collection with all intersecting items but exits continue even if tile is filled'
 );
+
+SELECT results_eq($$
+    select s from geojsonsearch('{"type": "Point","coordinates": [-87.75608539581299,30.692471153735646]}', '4688c09f64bc21b5e4bb7ce19c755623', '{"include":["id"]}'::jsonb, exitwhenfull => true, skipcovered => true) s;
+    $$,$$
+    select '{"type": "FeatureCollection", "features": [{"id": "pgstac-test-item-0097"}]}'::jsonb
+    $$,
+    'Test geojsonsearch to return feature collection with all intersecting items intersecting the point'
+);
```
vincentsarago, line 52: This should fail

bitner:
@vincentsarago this is super old ticket, but the entire point of the geometrysearch function is to deal with areas. If we are trying to just find the intersection with a point or line, we should just be using normal search with sortby and sending an intersects geometry. I'm not sure what the case here is for anything extra that the geometrysearch is providing when it isn't an area?

vincentsarago:
> geometrysearch function is to deal with areas

@bitner I could argue that `geometry` should work for any geometry 🤷‍♂️ 

It works great with `point` geometry but only if we force `exitwhenfull` and `skipcovered` to be false. 

https://github.com/stac-utils/titiler-pgstac/blob/main/titiler/pgstac/mosaic.py#L183-L199


I agree that normal `search` would be better but again I believed that using `geometrysearch` should have worked with all geometry types

bitner:
@vincentsarago I've added a check to the geometrysearch function that will set skipcovered and exitwhenfull to false whenever the input geometry is not a polygon. This passes the tests that you added (I just moved the tests to basic sql tests rather than pgtap tests to match the other existing tests).

bitner (approved):

Issue #53, opened 2021-09-14
allow additional `metadata` to be saved for searches row


vincentsarago:
https://github.com/stac-utils/pgstac/blob/b210b68dff7c561e1223546ef4e3d84649512682/sql/004_search.sql#L645-L655

PR #54, opened 2021-09-15
Searches metadata

vincentsarago:
Ref #53 

This PR adds:
- metadata column into `searches` table
- construct the search hash based on the metadata and the search body
- add metadata option in `search_query` function

### To Do
- [x] feedback from @bitner 
- [ ] create migration files
- [ ] add tests
- [ ] update changelog and prepare release

### DIFF
--- a/sql/004_search.sql
+++ b/sql/004_search.sql
@@ -651,19 +651,19 @@ CREATE TABLE IF NOT EXISTS searches(
     statslastupdated timestamptz,
     estimated_count bigint,
     total_count bigint,
-    metadata jsonb DEFAULT '{}'::jsonb
+    metadata jsonb DEFAULT NULL
 );
 
-CREATE OR REPLACE FUNCTION search_query(_search jsonb = '{}'::jsonb, updatestats boolean DEFAULT false, metadata jsonb = '{}'::jsonb) RETURNS searches AS $$
+CREATE OR REPLACE FUNCTION search_query(_search jsonb = '{}'::jsonb, updatestats boolean DEFAULT false, _metadata jsonb DEFAULT NULL) RETURNS searches AS $$
 DECLARE
     search searches%ROWTYPE;
 BEGIN
 INSERT INTO searches (search, metadata)
-    VALUES (search_tohash(_search), metadata)
+    VALUES (search_tohash(_search), _metadata)
     ON CONFLICT DO NOTHING
     RETURNING * INTO search;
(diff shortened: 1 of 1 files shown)


sql/004_search.sql
```diff
@@ -637,33 +637,33 @@ CREATE OR REPLACE FUNCTION search_tohash(jsonb) RETURNS jsonb AS $$
...
     search jsonb NOT NULL,
     _where text,
     orderby text,
     lastused timestamptz DEFAULT now(),
     usecount bigint DEFAULT 0,
     statslastupdated timestamptz,
     estimated_count bigint,
-    total_count bigint
+    total_count bigint,
+    metadata jsonb DEFAULT '{}'::jsonb
```
vincentsarago, line 654: maybe we don't have to set a default here (but then we should have a default in the `search_hash` function)

sql/004_search.sql
```diff
@@ -637,33 +637,33 @@ CREATE OR REPLACE FUNCTION search_tohash(jsonb) RETURNS jsonb AS $$
     SELECT $1 - '{token,limit,context,includes,excludes}'::text[];
 $$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
 
-CREATE OR REPLACE FUNCTION search_hash(jsonb) RETURNS text AS $$
-    SELECT md5(search_tohash($1)::text);
```
bitner, line 641: The reason we used the search_tohash function on the search json was so that we can remove the fields that aren't used to drive the context so that when doing things like paging or changing the returned fields, we can just use the cache from ie the previous page.

vincentsarago: I couldn't see where this function was used so I just decided to apply `search_toash` directly in the `search_hash` call https://github.com/stac-utils/pgstac/pull/54/files#diff-3c3e93de12c078e7f46e264188fcc2ba82eefd77be73795f964ad132d89af285R666

sql/004_search.sql
bitner: I think it's good to have a default and probably set it to not null.

That way we can merge update the metadata by using "UPDATE searches SET metadata = metadata || '{"newfield": 1}':::jsonb without having to do any coalesce or anything to deal with nulls.

vincentsarago: basically I set metadata to default to `{}` to make sure `search_hash` received two valid json object 🤷‍♂️

Issue #49, opened 2021-08-24
Multiple pgstac.migrations records can have same timestamp


kylebarron:
I was testing out pgstac migrations (with `pypgstac migrate`) and it looks like multiple version numbers can be inserted at the same time.

```
db=> SELECT * FROM pgstac.migrations;
 version |           datetime
---------+-------------------------------
 0.2.4   | 2021-08-24 18:34:59.364497+00
 0.2.7   | 2021-08-24 18:39:26.047286+00
 0.3.4   | 2021-08-24 18:39:26.124702+00
 0.2.8   | 2021-08-24 18:39:26.129415+00
 0.3.4   | 2021-08-24 18:39:26.133057+00
 0.2.9   | 2021-08-24 18:39:26.136371+00
 0.3.4   | 2021-08-24 18:39:26.136371+00
(7 rows)
```

`0.2.4` was added to the table manually, but all other rows were added as part of `pypgstac migrate`. This is an issue because `pypgstac migrate` picks up the most-recently-added row 
https://github.com/stac-utils/pgstac/blob/6e43fb3e691fb04376474ed122b5de15c55375d5/pypgstac/pypgstac/migrate.py#L113-L118

```
db=> SELECT version FROM pgstac.migrations
db-> ORDER BY datetime DESC LIMIT 1;
 version
---------
 0.2.9
(1 row)
```

If pgstac is dedicated to Semver, you could just take `Max(version)`?

```
db=> SELECT MAX(version) FROM pgstac.migrations;
  max
-------
 0.3.4
(1 row)
```

vincentsarago:
@bitner was this fixed in #57 ?

kylebarron:
I haven't tested this since #57 was merged and released but I suspect it might be fixed.

PR #57, opened 2021-09-16
Performance fixes

### LINKED ISSUE #51: Multiple entries of version in pgstc.migrations table
Running the pypgstac migrate command from a previous version to a new version leads to multiple entries of the final version in the pgstac.migrations table: 

```Migrating from 0.2.4 to 0.3.3.
Running migrations for ['pgstac.0.2.4-0.2.7.sql', 'pgstac.0.2.7-0.2.8.sql', 'pgstac.0.2.8-0.2.9.sql', 'pgstac.0.2.9-0.3.0.sql', 'pgstac.0.3.0-0.3.1.sql', 'pgstac.0.3.1-0.3.2.sql', 'pgstac.0.3.2-0.3.3.sql'].
pgstac version 0.3.3
```

```postgis=# select * from pgstac.migrations;
 version |           datetime
---------+-------------------------------
 0.2.4   | 2021-07-23 00:00:00+00
 0.2.7   | 2021-09-03 09:02:22.32407+00
 0.3.3   | 2021-09-03 09:02:22.360333+00
 0.2.8   | 2021-09-03 09:02:22.362009+00
 0.3.3   | 2021-09-03 09:02:22.362903+00
 0.2.9   | 2021-09-03 09:02:22.364059+00
 0.3.3   | 2021-09-03 09:02:22.364059+00
 0.3.0   | 2021-09-03 09:02:22.366378+00
 0.3.3   | 2021-09-03 09:02:22.366378+00
 0.3.1   | 2021-09-03 09:05:24.690869+00
 0.3.3   | 2021-09-03 09:05:24.690869+00
 0.3.2   | 2021-09-03 09:05:24.716236+00
 0.3.3   | 2021-09-03 09:05:24.716236+00
 0.3.3   | 2021-09-03 09:05:24.723922+00
 0.3.3   | 2021-09-03 09:05:24.723922+00
(15 rows)
```

For any migration file (e.g., 0.2.4 to 0.2.7) two inserts are currently conducted: 

1. From the migration file 
https://github.com/stac-utils/pgstac/blob/main/pypgstac/pypgstac/migrations/pgstac.0.2.4-0.2.7.sql#L313

2. From the python script (which inserts always to latest version to migrate to - in my case 0.3.3): 
https://github.com/stac-utils/pgstac/blob/main/pypgstac/pypgstac/migrate.py#L160-L170

I think this can be resolved by removing the insert statement from the python script?

### LINKED ISSUE #53: allow additional `metadata` to be saved for searches row

bitner:
- Add constraints to each partition to help query planner leverage constraint exclusion
  - Add constraints for collection_id and end_datetime
  - Add functionality to triggers to drop constraints before load and readd them after a load
  - Use explain with these constraints to get a list of partitions that would be hit by a query and use this list when looping through partitions
- Add metadata to searches table (fixes #53 superseded PR #54)
- Fixes for migration tagging (fixes #51 and #49)
- Adds knobs to control the use of context (instructions added in README on how to set)
  - pgstac.context setting options
    - 'on' always calculate the full count and report the context in the output
    - 'auto' use the pgstac.context_estimated_count and pgstac.context_estimated_count variables to determine whether to do the full accurate count or to report the estimated count in the context
    - 'off' do not report the context and never calculate the full count
   - pgstac.context_estimated_count -  number of estimated rows when in auto mode that when an estimated count is less than will trigger a full count
   - pgstac.context_estimated_cost  - estimated query cost from explain when in auto mode that when an estimated cost is less than will trigger a full count
   - pgstac.context_stats_ttl - an interval string ie "1 day" after which pgstac search will force recalculation of it's estimates

### DIFF
--- a/sql/004_search.sql
+++ b/sql/004_search.sql
@@ -38,7 +38,7 @@ IF cardinality(path_elements)<1 THEN
     path := field;
     path_txt := field;
     jsonpath := '$';
-    eq := NULL; -- format($F$ %s = %%s $F$, field);
+    eq := NULL;
     RETURN;
 END IF;
 
@@ -495,6 +495,23 @@ WITH t AS (
 FROM t;
 $$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
 
+CREATE OR REPLACE VIEW items_btree_indexes AS
+SELECT
+    substring(indexdef from 'btree (.*)')
+FROM pg_indexes
+WHERE
+    schemaname='pgstac'
+    AND tablename='items'
+;
+
+CREATE OR REPLACE FUNCTION field_orderby(p text) RETURNS text AS $$
+WITH t AS (
+    SELECT
+        replace(trim(substring(indexdef from 'btree \((.*)\)')),' ','')as s
+    FROM pg_indexes WHERE schemaname='pgstac' AND tablename='items' AND indexdef ~* 'btree'
+) SELECT s FROM t WHERE strpos(s, lower(trim(p)))>0;
+$$ LANGUAGE SQL;
+
 CREATE OR REPLACE FUNCTION sort_sqlorderby(
     _search jsonb DEFAULT NULL,
     reverse boolean DEFAULT FALSE
@@ -511,7 +528,7 @@ WITH sortby AS (
     SELECT jsonb_array_elements(sort) as value FROM withid
 ),sorts AS (
     SELECT
-        (items_path(value->>'field')).path as key,
+        coalesce(field_orderby((items_path(value->>'field')).path_txt), (items_path(value->>'field')).path) as key,
         parse_sort_dir(value->>'direction', reverse) as dir
     FROM withid_rows
 )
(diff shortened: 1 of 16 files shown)


vincentsarago (approved):
looks good as far as I understand 😄 

Should we add more tests for the new functions?

sql/004_search.sql
```diff
@@ -745,11 +877,16 @@ DECLARE
...
+    pend timestamptz;
+    pcurs refcursor;
+    search_where search_wheres%ROWTYPE;
 BEGIN
 searches := search_query(_search);
 _where := searches._where;
 orderby := searches.orderby;
-total_count := coalesce(searches.total_count, searches.estimated_count);
+search_where := where_stats(_where);
+total_count := coalesce(search_where.total_count, search_where.estimated_count);
```
lossyrob, line 889: Is total_count calculated even if context == 'off'? Or is this lazily evaluated and doesn't perform it's operation until the jsonb_build_object call below?

bitner: total_count and estimated_count are both calculated in the where_stats function. estimated_count is always calculated (it's a free side effect of what we use to get the list of partitions) and total_count is not calculated in there if the context is set to 'off'. The coalesce here takes the first non-null argument, so it will use the total if it was either calculated or existed already in the cache table.

lossyrob (approved):

PR #70, opened 2021-11-16
Cql2

### LINKED ISSUE #67: Support for CQL2
cql2 docs:

- https://github.com/radiantearth/stac-api-spec/pull/225 
- https://github.com/opengeospatial/ogcapi-features/issues/601#issuecomment-940066783 
- https://github.com/radiantearth/stac-api-spec/blob/f5da775080ff3ff46d454c2888b6e796ee956faf/fragments/filter/README.md

### LINKED ISSUE #69: Support isNull in filters
[As described here](https://github.com/radiantearth/stac-api-spec/tree/master/fragments/filter#example-9-using-is-null), it would be useful to be able to filter a search request based on whether a given property is null.

bitner:
- Bump version to 0.4.0
- Fixes syntax for IN, BETWEEN, ISNULL, and NOT in CQL 1 (fixes #69)
- Adds support for modifying settings through pgstac_settings table and by passing in 'conf' object in search json to support AWS RDS where custom user configuration settings are not allowed and changing settings on the fly for a given query.
- Adds support for CQL2-JSON (fixes #67)
  - Adds tests for all examples in https://github.com/radiantearth/stac-api-spec/blob/f5da775080ff3ff46d454c2888b6e796ee956faf/fragments/filter/README.md
  - filter-lang parameter controls which dialect of CQL to use
    - Adds 'default-filter-lang' setting to control what dialect to use when 'filter-lang' is not present
    - old style stac 'query' object and top level ids, collections, datetime, bbox, and intersects parameters are only available with cql-json


sql/004_search.sql
```diff
@@ -441,25 +472,166 @@ END;
...
+        "gte": "%s >= %s",
+        "le": "%s <= %s",
+        "ge": "%s >= %s",
+        "=": "%s = %s",
+        "<": "%s < %s",
+        "<=": "%s <= %s",
+        ">": "%s > %s",
+        ">=": "%s >= %s",
+        "like": "%s LIKE %s",
+        "ilike": "%s ILIKE %s",
```
mmcfarland, line 506: Curious if this is CQL2 oriented, using the CQL1 spec, it looked like the [op `LIKE` was used with a `nocase` flag](https://portal.ogc.org/files/96288#:~:text=J-,If%20NOCASE%20is%20TRUE,-%2C%20the%20comparison%20SHALL). If that's not hte case anymore, could you add a test that indicates correct usage (as expected by pgStac)?

mmcfarland (commented):
Looks good, @bitner. I had a question about case insensitive 'LIKE' queries and suggested a test for it and the `BETWEEN` that was fixed in this PR.

lossyrob (approved):

sql/004_search.sql
bitner: cql2 uses upper and lower pseudo function on the property. Examples added in the tests.

PR #78, opened 2022-01-03
Work v0.4.1

### LINKED ISSUE #71: Intersects geometry in search appears limited to 8K bytes in version >= 0.3.5
When running searches that include an "intersects" for complex geometries, I am seeing this error:

`ERROR:  index row requires 293792 bytes, maximum size is 8191`

The "293792 bytes" seems to correspond to the size of the input geometry. 

The error does not occur for simpler geometries (I ran one that was a simple polygon covering the same area loosely as the complex polygon) nor in earlier versions of pgstac. Below is the stack trace from the database when running a search with one collection and an intersects of California, although I left off the first two notices of `NOTICE:  ARGS after array cleaning: ` since the geometry value is so long. Happy to provide the specific California geometry if needed.

```
NOTICE:  Checking if update is needed.
NOTICE:  Stats Last Updated: <NULL>
NOTICE:  TTL: 1 day, Age: <NULL>
NOTICE:  Context: off, Existing Total: <NULL>
NOTICE:  Time for just the explain: 00:00:00.428574
NOTICE:  Time for explain + join: 00:00:00.431062

ERROR:  index row requires 293792 bytes, maximum size is 8191
CONTEXT:  SQL statement "INSERT INTO search_wheres SELECT sw.*

    ON CONFLICT (_where)

    DO UPDATE

        SET

            lastused = sw.lastused,

            usecount = sw.usecount,

            statslastupdated = sw.statslastupdated,

            estimated_count = sw.estimated_count,

            estimated_cost = sw.estimated_cost,

            time_to_estimate = sw.time_to_estimate,

            partitions = sw.partitions,

            total_count = sw.total_count,

            time_to_count = sw.time_to_count"
PL/pgSQL function where_stats(text,boolean) line 189 at SQL statement
SQL statement "SELECT where_stats(search._where, updatestats)"
PL/pgSQL function search_query(jsonb,boolean,jsonb) line 41 at PERFORM
PL/pgSQL function search(jsonb) line 71 at assignment
SQL state: 54000
```
Thank you!

### LINKED ISSUE #73: Update Changelog!
Changelog hasn't been updated since 0.3.4 😬

### LINKED ISSUE #77: Can't override `default-filter-lang` if it is set to `cql-json`
If the `pgstac_settings` `default-filter-lang` is set to `cql-json`, I don't believe it can be overridden as intended by this line:

https://github.com/stac-utils/pgstac/blob/3afa5df36c0a54cf62339d25e7ce23320848e70f/sql/004_search.sql#L651

In the case where the default is set to `"cql-json"` and the client sends a `filter-lang` property of `"cql2-json"`, I read the logic of that lines as  `(false or true)` and it then attempts and fails to parse the cql2 as cql (which is the behavior I'm seeing). Instead, I think the default just need to be checked is the search didn't provide a filter-lang property.

If the default is set to `cql2-json`, the search-supplied and override behavior seems correct.

bitner:
- Bumps version to 0.4.1
- Adds changelog for v0.4.0 and v0.4.1 (fixes #73)
- Adds Fix for large queries in the query cache (fixes #71)
- Adds Fix to make sure that filter-lang always trumps default system filter settings (fixes #77)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,29 @@
 # Changelog
 
-## Ureleased
+## [v0.4.1]
 
-## Changed
+### Changed
 
 - Update `typer` to 0.4.0 to avoid clashes with `click` ([#76](https://github.com/stac-utils/pgstac/pull/76))
+
+### Fixed
+
+- Fix logic in getting settings to make sure that filter-lang set on query is respected. ([#77](https://github.com/stac-utils/pgstac/pull/77))
+- Fix for large queries in the query cache. ([#71](https://github.com/stac-utils/pgstac/pull/71))
+
+
+## [v0.4.0]
+
+### Fixed
+- Fixes syntax for IN, BETWEEN, ISNULL, and NOT in CQL 1 ([#69](https://github.com/stac-utils/pgstac/pull/69))
+
+### Added
+- Adds support for modifying settings through pgstac_settings table and by passing in 'conf' object in search json to support AWS RDS where custom user configuration settings are not allowed and changing settings on the fly for a given query.
+- Adds support for CQL2-JSON ([#67](https://github.com/stac-utils/pgstac/pull/67))
+  - Adds tests for all examples in https://github.com/radiantearth/stac-api-spec/blob/f5da775080ff3ff46d454c2888b6e796ee956faf/fragments/filter/README.md
+  - filter-lang parameter controls which dialect of CQL to use
+    - Adds 'default-filter-lang' setting to control what dialect to use when 'filter-lang' is not present
+    - old style stac 'query' object and top level ids, collections, datetime, bbox, and intersects parameters are only available with cql-json
 
 ## [v0.3.4]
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,30 @@
...
+- Fixes syntax for IN, BETWEEN, ISNULL, and NOT in CQL 1 ([#69](https://github.com/stac-utils/pgstac/pull/69))
+
+### Added
+- Adds support for modifying settings through pgstac_settings table and by passing in 'conf' object in search json to support AWS RDS where custom user configuration settings are not allowed and changing settings on the fly for a given query.
+- Adds support for CQL2-JSON ([#67](https://github.com/stac-utils/pgstac/pull/67))
+  - Adds tests for all examples in https://github.com/radiantearth/stac-api-spec/blob/f5da775080ff3ff46d454c2888b6e796ee956faf/fragments/filter/README.md
+  - filter-lang parameter controls which dialect of CQL to use
+    - Adds 'default-filter-lang' setting to control what dialect to use when 'filter-lang' is not present
+    - old style stac 'query' object and top level ids, collections, datetime, bbox, and intersects parameters are only available with cql-json
+
```
vincentsarago, line 27: ❤️

vincentsarago (approved):

PR #79, opened 2022-01-05
Added details for updating pgstac settings to readme

zstatmanweil:
Hi @bitner , thanks for all the recent work on this! The `pgstac_settings` is a really useful change in 0.4.0 so I added it to the README so it would be obvious to others.


bitner (approved):
Thanks @zstatmanweil! I'll merge this into master and include it with the next release.

PR #81, opened 2022-01-13
fix for optimization when operator is '='

bitner:
- Fix for the equals optimization when querying the properties json. Prior to this, it would only kick in when the operator was "eq", now it will kick in for both "eq" and "=".
- Update version to 0.4.3


mmcfarland (approved):
Looks great, thanks!

PR #87, opened 2022-01-24
remove useless asyncio requirement

### LINKED ISSUE #86: remove asyncio from pypgstac requirements
https://github.com/stac-utils/pgstac/blob/4b582773f764f317788a8879f5cac3ee4f4f7951/pypgstac/pyproject.toml#L18

AsyncIO is part native python install so there is no need to force-install it.

This could cause issue when using pypgstac in AWS Lambda: https://stackoverflow.com/a/61746719

vincentsarago:
closes #86

### DIFF
--- a/pypgstac/pyproject.toml
+++ b/pypgstac/pyproject.toml
@@ -15,7 +15,6 @@ python = ">=3.7"
 asyncpg = "^0.22.0"
 smart-open = "^4.2.0"
 typer = ">=0.4.0"
-asyncio = "^3.4.3"
 orjson = "^3.5.2"
 python-dateutil = "^2.8.2"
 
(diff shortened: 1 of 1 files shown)


pypgstac/pyproject.toml
```diff
@@ -15,7 +15,6 @@ python = ">=3.7"
 asyncpg = "^0.22.0"
 smart-open = "^4.2.0"
 typer = ">=0.4.0"
-asyncio = "^3.4.3"
 orjson = "^3.5.2"
```
vincentsarago, line 18: @bitner do we really need to `hard` pin orjson to ~`3.5.*`?

bitner (approved):
Thanks @vincentsarago you can remove the hard pin on orjson too.

pypgstac/pyproject.toml
vincentsarago, line 18:
```suggestion
orjson = ">=3.5.2"
```

PR #93, opened 2022-03-02
[TO BE DISCUSSED] add type to search entry

vincentsarago:
context: https://github.com/developmentseed/eoAPI/issues/22, https://github.com/stac-utils/titiler-pgstac/pull/45

In Titiler-PgSTAC (and eoAPI) we are exploring ways to search/list search entries (for mosaic)


bitner:
@vincentsarago since a query can be both a search and a mosaic, I think it would be sufficient to just look for the presence of they type='mosaic' attribute on the metadata. For that, rather than adding another column to track the type, we could just build an index on that json item

```
CREATE INDEX ON searches ((true)) WHERE metadata->>'type'='mosaic';
```

With that, the index size should be minimal since it is only on things that are "mosaics" and could be very quickly accessed using 
```
SELECT * FROM searches WHERE metadata->>'type'='mosaic';
```

vincentsarago:
👍 I wasn't sure if we could create an index on something that might not exists 🤷

vincentsarago:
I'm going to closes this for now because I don't want to add something unrelated to the core PGStac! (I'll add specific documentation in eoAPI or titiler-pgstac)

PR #96, opened 2022-03-22
v0.5.0

bitner:
## [v0.5.0]
Version 0.5.0 is a major refactor of how data is stored. It is recommended to start a new database from scratch and to move data over rather than to use the inbuilt migration which will be very slow for larger amounts of data.

### Changed
 - The partition layout has been changed from being hardcoded to a partition to week to using nested partitions. The first level is by collection, for each collection, there is an attribute partition_trunc which can be set to NULL (no temporal partitions), month, or year.

 - CQL1 and Query Code have been refactored to translate to CQL2 to reduce duplicated code in query parsing.

 - Unused functions have been stripped from the project.

 - Pypgstac has been changed to use Fire rather than Typer.

 - Pypgstac has been changed to use Psycopg3 rather than Asyncpg to enable easier use as both sync and async.

 - Indexing has been reworked to eliminate indexes that from logs were not being used. The global json index on properties has been removed. Indexes on individual properties can be added either globally or per collection using the new queryables table.

 - Triggers for maintaining partitions have been updated to reduce lock contention and to reflect the new data layout.

 - The data pager which optimizes "order by datetime" searches has been updated to get time periods from the new partition layout and partition metadata.

 - Tests have been updated to reflect the many changes.

### Added

 - On ingest, the content in an item is compared to the metadata available at the collection level and duplicate information is stripped out (this is primarily data in the item_assets property). Logic is added in to merge this data back in on data usage.


Dockerfile.dev
```diff
@@ -12,7 +12,9 @@ ENV \
 
 RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
 
-RUN pip install poetry==1.1.7
+RUN pip install --upgrade pip && \
+    pip install --upgrade poetry==1.1.13 && \
+    pip install --upgrade psycopg[binary]
```
vincentsarago, line 17: might be needed to add `pool` here too

```
psycopg[binary,pool]
```

Dockerfile
```diff
@@ -33,8 +33,9 @@ RUN \
         python3-setuptools \
     && pip3 install -U pip setuptools packaging \
     && pip3 install -U psycopg2-binary \
```
vincentsarago, line 35: do we still need `psycopg2` ?

bitner: Yeah, it's used by the diff tool that helps generate migrations.

pypgstac/pypgstac/db.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+@dataclass
+class settings:
+    db_min_conn_size = 0
+    db_max_conn_size = 1
+    db_max_queries = 5
+    db_max_idle = 5
+    db_num_workers = 1
+    db_retries = 3
```
vincentsarago, line 26: can we use pydantic BaseSettings here?

Dockerfile
vincentsarago: isn't it risky to have both psycopg2 and psycopg3 ?

Dockerfile.dev
bitner: psycopg_pool should get pulled from the poetry config. It's just psycopg itself that has issues there.

Dockerfile
bitner: I don't see why? The modules are all very clearly psycopg or psycopg2. But I'll take a look at pgmigra to see if there have been any updates, it connects through sqlalchemy, so it's possible it could use psycopg3 now.

bitner: I could always just run migra from it's own docker container too to keep it entirely separate.

Dockerfile.dev
lossyrob: @bitner  Can `psycopg[binary]` be declared as a dependency in the published package? I was installing this pypgstac from source but running into:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/psycopg/__init__.py", line 9, in <module>
    from . import pq  # noqa: F401 import early to stabilize side effects
  File "/usr/local/lib/python3.8/dist-packages/psycopg/pq/__init__.py", line 114, in <module>
    import_from_libpq()
  File "/usr/local/lib/python3.8/dist-packages/psycopg/pq/__init__.py", line 106, in import_from_libpq
    raise ImportError(
ImportError: no pq wrapper available.
Attempts made:
- couldn't import psycopg 'c' implementation: No module named 'psycopg_c'
- couldn't import psycopg 'binary' implementation: No module named 'psycopg_binary'
- couldn't import psycopg 'python' implementation: libpq library not found
```

I had to duplicate dependencies as I don't think the pip install from git likes the poetry.toml, but found I had to use `"psycopg[binary]==3.0.10",` to get past these errors. Does poetry support dependencies with extras?

pypgstac/pypgstac/load.py
```diff
@@ -1,246 +1,501 @@
...
+                logging.info("File cannot be read as json")
+                raise
+
+
+@dataclass
+class Loader:
+    """Utilities for loading data."""
+
+    db: PgstacDB
+    minimizeproj: Optional[bool] = False
```
lossyrob, line 165: Is this used?

pypgstac/pypgstac/db.py
```diff
@@ -0,0 +1,251 @@
...
+                kwargs={
+                    "options": "-c search_path=pgstac,public"
+                    " -c application_name=pypgstac"
+                },
+            )
+        return self.pool
+
+    def open(self) -> None:
+        """Open database pool connection."""
+        print("open")
```
lossyrob, line 94: Note to remove debug prints before merge

pypgstac/pypgstac/load.py
```diff
@@ -1,246 +1,501 @@
...
+                        "Available modes are insert, ignore, upsert, and delsert."
+                        f"You entered {insert_mode}."
+                    )
+        logging.debug(
+            f"Copying data for {partition} took {time.perf_counter() - t} seconds"
+        )
 
+    def load_items(
+        self,
+        file: Optional[str] = "stdin",
```
lossyrob, line 386: This loses the ability to pass in an `Iterator[Dict[str, Any]]` or `Iterator[bytes]` which we were using. Can we make this loader similarly work with things other than files?

pypgstac/pypgstac/db.py
```diff
@@ -0,0 +1,251 @@
...
+            self.pool.close()
+
+    def connect(self) -> Connection:
+        """Return database connection."""
+        pool = self.get_pool()
+        if self.connection is None:
+            self.connection = pool.getconn()
+            if self.debug:
+                self.connection.add_notice_handler(pg_notice_handler)
+        atexit.register(self.disconnect)
```
lossyrob, line 110: This should tab in; if self.connection is not None then the DB is already connected, so another exit handler shouldn't be registered.

pypgstac/pypgstac/load.py
bitner: Ah, I stripped that out. When I was testing everything I could use to compress, I did have code that would look up the EPSG code if something had a proj:wkt_text and swap it out with just the reference to the code. I'll get rid of the flag.

lossyrob (approved):

PR #100, opened 2022-04-06
Add LICENSE

### LINKED ISSUE #99: No license specified
We're using PGSTAC at our company, and I'm trying to fill in license information for internal approval.

I've found that there is no license information specified in a `LICENSE` file or in the Github metadata.

What license is this code released under?

(If you haven't assigned a license yet, then a lot of the rest of the stac-utils seem to be released under the MIT license)

lossyrob:
Add MIT license for this project.

closes #99

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+    MIT License
+
+    Copyright (c) Microsoft Corporation and other contributors.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+    MIT License
+
+    Copyright (c) Microsoft Corporation and other contributors.
```
lossyrob, line 3: @bitner To be explicit about thinking here: MIT is standard license for Microsoft open source, and this spot on the license file usually lists the initial contributor; since this work was kicked off for Microsoft this seems the most appropriate listing.

vincentsarago, line 3: is this really a Microsoft Corp project or should we use `stac-utils` 🤷‍♂️ 

just looking at all other projects in stac-utils org, we have a tons of differents `org`

lossyrob: stac-utils isn't a copyright holder; see above comment. Standard practice (in my experience) is to list the original holder and say others, or list copyright holders specifically. Technically every contributor retains the copyright of their contributions (or the companies who were paying for that code to be developed). Copyright doesn't get signed over to anyone else unless there's a CLA that explicitly states that (like the Apache CLA, I believe)

PR #106, opened 2022-04-28
Bug fixes for 0.5.1

### LINKED ISSUE #109: Can't delete collections due to foreign key constraint on `partitions` table
If this is an issue with how I've setup my partitions (currently I _haven't_ done any tweaking past what ever is done by default) please let me know, but if believe this might be an issue with `v0.5.1`?

---

## Reproduce:
- Have a collection within pgstac
- Either:
  - Use the `delete_collection` function directly

image

  - Interface indirectly with `delete_collection` by reaching out to `DELETE /collections/{collection}`

image

Both results in a similar error from Postgres:

image

bitner:
- Bump version to 0.5.2

### Loader
- Add "--dehydrated" option to loader to be able to load a dehydrated file (or iterable) of items such as would be output using pg_dump or postgresql copy.
- Add "--chunksize" option to loader that can split the processing of an iterable or file into chunks of n records at a time

### Hydration
- Update hydration/dehydration logic to make sure that it matches hydration/dehydration in pypgstac
- Add tests to ensure that pgstac and pypgstac hydration logic is equivalent

### Fields
- Update fields logic in pgstac to only use full paths and to match logic in stac-fastapi

### Fixes
- Fix function signatures for transactional functions (delete_item etc) to make sure that they are marked as volatile
- Fix function for getting start/end dates from a stac item
- Add triggers and on delete cascade for collections to clean up partitions when removing a collection (fixes #109 )


lossyrob:
@bitner updated this to main

lossyrob (approved):

mmcfarland (approved):

PR #107, opened 2022-05-03
Add hydration and dehydration methods loading and nohydrate mode

mmcfarland:
* Converts poetry to setup.py module definition.
* Adds hydration method to be used to restore items to their full value against a base_item in Python. This is used when the `conf` value `nohydrate` is used within a pgstac search.
* Refactors the existing `format_item` into a dehydrate method that is compatible with the new hydrate method.
* Adds tests for the specific hydration rules


pypgstac/setup.py
```diff
@@ -0,0 +1,55 @@
...
+with open("README.md") as f:
+    desc = f.read()
+
+install_requires = [
+    "smart-open==4.2.*",
+    "orjson>=3.5.2",
+    "python-dateutil==2.8.*",
+    "fire==0.4.*",
+    "psycopg[binary]==3.0.*",
+    "psycopg-pool==3.1.*",
```
mmcfarland, line 14: I think how we handle these deps with the plan of making pypgstac a dependency of stac-fastapi.pgstac is still outstanding.

Dockerfile
```diff
@@ -49,12 +46,9 @@ RUN mkdir -p /opt/src/pypgstac
...
 WORKDIR /opt/src/pypgstac
 
-COPY pypgstac/poetry.lock pypgstac/pyproject.toml ./
-RUN poetry install
-
-
 COPY pypgstac /opt/src/pypgstac
-RUN poetry install
+
+RUN pip3 install -e /opt/src/pypgstac
```
TomAugspurger, line 51:
```suggestion
RUN python3 -m pip install -e /opt/src/pypgstac
```

*might* be the cause of the error you're seeing in CI.

Dockerfile.dev
```diff
@@ -6,26 +6,20 @@ ENV \
...
 
 WORKDIR /opt/src/pypgstac
 
-COPY pypgstac/poetry.lock pypgstac/pyproject.toml ./
-RUN poetry install
-
 
 COPY pypgstac /opt/src/pypgstac
-RUN poetry install
+RUN pip3 install -e /opt/src/pypgstac[dev]
```
bitner, line 22: We should do this same as in Dockerfile with python3 -m pip install...

bitner:
scripts/bin/stageversion is going to need to be updated to not use poetry

lossyrob:
Fixed CI. This PR LGTM - @mmcfarland @bitner ready for your review

lossyrob (approved):

mmcfarland (commented):
Can't hit approve on my own PR, but looks good to me ✅

PR #110, opened 2022-05-11
Add autocommit to connections

lossyrob:
This PR adds `.autocommit = True` to new connections from `PgstacDB`. 

I ran into an issue where long, multi-load ingests were holding locks for the duration of the entire ingest. There was a transaction was open the entire time the connection was open, and so the partition being ingested into was lockd the entire ingest. Using autocommit scoped the lock issues to a single scoped call to `load_items`.

Based on the [Hint at the bottom of these docs](https://www.psycopg.org/psycopg3/docs/basic/transactions.html?msclkid=496fe3b1d0d511eca8f7aae193c127a1#transaction-contexts), `autocommit` on the connections that are established in PgstacDB to avoid any transaction issues.

In addition, a scoped `logger` is used so that it's possible for client code to adjust logging levels of pypgstac directly, e.g. turn on debug logging just for the package.

### DIFF
--- a/pypgstac/pypgstac/db.py
+++ b/pypgstac/pypgstac/db.py
@@ -103,6 +103,7 @@ class PgstacDB:
         pool = self.get_pool()
         if self.connection is None:
             self.connection = pool.getconn()
+            self.connection.autocommit = True
             if self.debug:
                 self.connection.add_notice_handler(pg_notice_handler)
             atexit.register(self.disconnect)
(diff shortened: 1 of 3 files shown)


pypgstac/pypgstac/db.py
```diff
@@ -103,6 +105,7 @@ def connect(self) -> Connection:
         pool = self.get_pool()
         if self.connection is None:
             self.connection = pool.getconn()
+            self.connection.autocommit = True
```
lossyrob, line 108: Autocommit change is here.

bitner (approved):

PR #112, opened 2022-05-13
Ensure special keys are not in content when loaded

mmcfarland:
The loader pop'd keys like `collection` off the item, then dehydrated the
item to be used as the 'content'. With `collection` removed prior to
dehydration, it was flagged with the "do-not-merge" marker because the
key is on the base_item. Instead, ensure that id, collection, and
geometry are not in `content` as they are stored on the table row and
shouldn't participate in hydration.

Additionally, bbox was previously a derived value at search runtime but
was recently changed to a returned value if it existed on the item.
However, during loading, the bbox was dropped so would never exist on
the persisted item.

### DIFF
--- a/pypgstac/pypgstac/load.py
+++ b/pypgstac/pypgstac/load.py
@@ -544,9 +544,8 @@ class Loader:
 
         out["partition"] = partition
 
-        bbox = item.get("bbox")
         geojson = item.get("geometry")
-        if geojson is None and bbox is not None:
+        if geojson is None:
             geometry = None
         else:
             geometry = str(Geometry.from_geojson(geojson).wkb)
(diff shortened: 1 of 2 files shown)


mmcfarland:
cc @lossyrob @bitner

pypgstac/pypgstac/load.py
```diff
@@ -544,8 +544,8 @@ def format_item(self, _item: Union[Path, str, dict]) -> dict:
 
         out["partition"] = partition
 
-        bbox = item.pop("bbox")
-        geojson = item.pop("geometry")
+        bbox = item.get("bbox")
```
lossyrob, line 547: Is `bbox` used? If not, this line can just be removed

lossyrob: Actually I see where it's used now - but I'm not sure I follow the logic. Wouldn't `geojson is None and bbox is not None` fail if the bbox check caused the statement to evaluate to False?

lossyrob: Discussed in call, this is being removed and the logic below being changed

lossyrob (approved):

PR #114, opened 2022-05-14
Avoid updating partitions table when unnecessary

lossyrob:
This commit refactors the loader code to avoid unnecessary partition updates. Partition updates should be avoided when unnecessary as they can have a large performance impact for partitions containing many items.

### DIFF
--- a/pypgstac/pypgstac/load.py
+++ b/pypgstac/pypgstac/load.py
@@ -1,5 +1,6 @@
 """Utilities to bulk load data into pgstac from json/ndjson."""
 import contextlib
+from datetime import datetime
 import itertools
 import logging
 from pathlib import Path
@@ -25,6 +26,7 @@ import psycopg
 from orjson import JSONDecodeError
 from plpygis.geometry import Geometry
 from psycopg import sql
+from psycopg.types.range import Range
 from smart_open import open
 from tenacity import (
     retry,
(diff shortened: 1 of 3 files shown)


pypgstac/pypgstac/load.py
```diff
@@ -137,15 +148,18 @@ def read_json(file: Union[Path, str, Iterator[Any]] = "stdin") -> Iterable:
                 yield orjson.loads(line)
 
 
-@dataclass
```
lossyrob, line 140: Remove dataclass; this is more appropriate as a normal Python object. IMO dataclasses should have light functionality and really be kept to data containers with public members (sort of like a struct). The dataclass requirement for immutable defaults was forcing the need to make this private attribute optional, which triggered the change from a dataclass to a standard object.

pypgstac/pypgstac/load.py
```diff
@@ -548,7 +572,10 @@ def format_item(self, _item: Union[Path, str, dict]) -> dict:
         if geojson is None:
             geometry = None
         else:
-            geometry = str(Geometry.from_geojson(geojson).wkb)
+            geom = Geometry.from_geojson(geojson)
+            if geom is None:
+                raise Exception(f"Invalid geometry encountered: {geojson}")
+            geometry = str(geom.wkb)
```
lossyrob, line 578: Check for None to appease type linter

lossyrob:
Tested this on a re-ingest of Sentinel 1 GRD with about a ~10x improvement in ingest timing

PR #121, opened 2022-06-03
Fix loading issues with dehydrated load

lossyrob:
This moves away from using `csv` to read the dehydrated, direct table output and fixes an issue with loading quoted values.

The CSV reader has a field size limit that can be reached with large items, producing:

```
_csv.Error: field larger than field limit (131072)
```

This can be worked around by setting a larger value through a method like:

```python
def set_csv_field_size_limit() -> None:
    """Set the field size limit to the maximum allowed.
    Otherwise csv can raise _csv.Error: field larger than field limit (131072)
    """
    field_size_limit = sys.maxsize
    while True:
        try:
            csv.field_size_limit(field_size_limit)
            break
        except OverflowError:
            field_size_limit = int(field_size_limit / 10)
```

However, the fact we use the CSV reader unquoted tab-delimited also seems sus - what happens if an Item description has a tab in it? This PR contains changes that moves away from using the csv reader and ensures this wouldn't be an issue.

Also, loading a dumped item table from the PC found another issue - items with escape quoted values (e.g. WKT2) were causing errors in loading. This replaces the troublesome quote character on reading in the table data.


bitner (changes requested):
It looks like we need to add the test data for the pypgstac tests to get CI passing. I't bombing on loading chloris.json as a collection. Once the tests are passing I am :+1: to merge.

Issue #123, opened 2022-06-08
pypgstac and pgstac database compatibility


drnextgis:
I have a doubt regarding compatibility of pypgstac and pgstac database with different versions.

I use pypgstac 0.4.3 to upload data to the pgstac database programmatically:

```python
import asyncio
from pypgstac.load import load_ndjson

dsn = "..."
asyncio.run(
    load_ndjson(file="items.njson", table="items", dsn=dsn, method="insert")
)
```

Is it safe to use this code to populate v0.6.X pgstac database or pypgstac must be the same version as the database?

I think it makes sense to reflect this information in docs.

bitner:
Yes, pypgstac and pgstac should be used with the same versions. We should definitely add this to the documentation and probably add a check to the loader to make sure that the loader is compatible with the target database.

bitner:
Just time to review it. I'm preparing a few items for a release this week, I'll make sure to review this to try to get it in the release.

PR #125, opened 2022-06-13
Add a check to the loader to make sure that the loader is compatible with the target database

drnextgis:
Fixes https://github.com/stac-utils/pgstac/issues/123


bitner (approved):
Looks good. Thanks @drnextgis!

PR #127, opened 2022-06-21
Add support for array operators

bitner:
Adds support for Array Operators (aequals, acontains, contained by, aoverlaps).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # Changelog
+## [v0.6.6]
+### Added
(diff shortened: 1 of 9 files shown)


bitner:
Hrmmm, looking at the different OGC documents now, there are some places that use aequals, aoverlaps, acontains, and contained by and others that use a_equals, a_overlaps, a_contains and a_contained_by. The latter is definitely more consistent with spatial and temporal operators.

lossyrob (approved):
As discussed on a call, I suggest making the test coverage robust, but otherwise +1

mmcfarland (approved):

PR #133, opened 2022-07-05
Queryables

bitner:
### Added
- Add get_queryables function to return a composite queryables json for either a single collection (text), a list of collections(text[]), or for the full repository (null::text).

- Add missing_queryables(collection text, tablesample int) function to help identify if there are any properties in a collection without entries in the queryables table. The tablesample parameter is an int <=100 that is the approximate percentage of the collection to scan to look for missing queryables rather than reading every item.

- Add missing_queryables(tablesample int) function that scans all collections using a sample of records to identify missing queryables.

### DIFF
--- a/sql/002a_queryables.sql
+++ b/sql/002a_queryables.sql
@@ -13,10 +13,11 @@ CREATE INDEX queryables_property_wrapper_idx ON queryables (property_wrapper);
 
 INSERT INTO queryables (name, definition) VALUES
 ('id', '{"title": "Item ID","description": "Item identifier","$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json#/definitions/core/allOf/2/properties/id"}'),
-('datetime','{"description": "Datetime","type": "string","title": "Acquired","format": "date-time","pattern": "(\\+00:00|Z)$"}')
+('datetime','{"description": "Datetime","type": "string","title": "Acquired","format": "date-time","pattern": "(\\+00:00|Z)$"}'),
+('geometry', '{"title": "Item Geometry","description": "Item Geometry","$ref": "https://geojson.org/schema/Feature.json"}')
 ON CONFLICT DO NOTHING;
 
-
+ON CONFLICT DO NOTHING;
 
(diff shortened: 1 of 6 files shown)


sql/002a_queryables.sql
```diff
@@ -13,11 +13,10 @@ CREATE INDEX queryables_property_wrapper_idx ON queryables (property_wrapper);
 
 INSERT INTO queryables (name, definition) VALUES
 ('id', '{"title": "Item ID","description": "Item identifier","$ref": "https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json#/definitions/core/allOf/2/properties/id"}'),
-('datetime','{"description": "Datetime","type": "string","title": "Acquired","format": "date-time","pattern": "(\\+00:00|Z)$"}')
+('datetime','{"description": "Datetime","type": "string","title": "Acquired","format": "date-time","pattern": "(\\+00:00|Z)$"}'),
```
mmcfarland, line 16: I'm a little curious what should be advertised when a collection's item actually have `start_datetime` and `end_datetime` defined, instead of `datetime`. Should they be rolled up into just this `datetime` entry, or should that be a part of the sampling? At least in pgstac, querying against datetime is still supported even when the field is not populated.

https://github.com/radiantearth/stac-spec/blob/master/item-spec/json-schema/datetime.json

mmcfarland (approved):
Looks good, I was able to run this in a staging DB and generate queryable attributes.

sql/002a_queryables.sql
bitner: In pgstac, the instantiated columns datetime and end_datetime are always populated and cannot be null. If the properties "datetime" is set, the datetime and end_datetime columns will both equal properties.datetime. If properties.datetime does not exist or is null, datetime=properties.start_datetime and end_datetime=properties.end_datetime.

From a querying perspective, if you are using datetime=start/end or any of the temporal operators with {"property":"datetime"}, pgstac is comparing against the time range generated from the datetime/end_datetime instantiated columns. If using regular operators and you specify {"property":"start_datetime"} or {"property":"end_datetime"} it will query against those values in the properties json as individual values (not as a time range). So they are still nominally accessible if you were to query them directly.

PR #142, opened 2022-09-22
update pinned versions of several dependencies

hrodmn:
I am working on incorporating `pgstac` into a project and I had to downgrade to smart-open==4.2.0 so I thought I would try bumping the pinned version to the latest (6.2.*). Since I was in there already I bumped a few other less-outdated pinned versions, too. All of the tests are passing when I run `scripts/test` as described in the contributing guidelines!

Maybe there is a good reason smart-open is held back to version 4.2.* but in case there is not one, here is a PR!

### DIFF
--- a/pypgstac/setup.py
+++ b/pypgstac/setup.py
@@ -6,7 +6,7 @@ with open("README.md") as f:
     desc = f.read()
 
 install_requires = [
-    "smart-open==6.2.*",
+    "smart-open>=4.2,<7.0",
     "orjson>=3.5.2",
     "python-dateutil==2.8.*",
     "fire==0.4.*",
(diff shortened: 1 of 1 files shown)


pypgstac/setup.py
```diff
@@ -6,13 +6,13 @@
     desc = f.read()
 
 install_requires = [
-    "smart-open==4.2.*",
+    "smart-open==6.2.*",
```
vincentsarago, line 9: Hard requirements can be dangerous, IMO we should try to have versions opened to a specific major release (e.g `>=6.0,<7.0`)

vincentsarago: but if it worked with 4.2 we could do `>=4.2,<7.0`

hrodmn: Good idea! I will update the strict requirements to be more permissive.

vincentsarago (approved):

hrodmn:
@vincentsarago I'm sorry I left this hanging - thanks for taking this over the finish line!

PR #145, opened 2022-10-25
Don't return valid queryable object for invalid collection 

mmcfarland:
Currently, passing in a non-existent collection will return a valid queryable object.

### DIFF
--- a/pypgstac/pypgstac/migrations/pgstac.0.6.9-0.6.10.sql
+++ b/pypgstac/pypgstac/migrations/pgstac.0.6.9-0.6.10.sql
@@ -8,9 +8,13 @@ CREATE OR REPLACE FUNCTION pgstac.get_queryables(_collection_ids text[] DEFAULT 
 AS $function$
 BEGIN
     -- Build up queryables if the input contains valid collection ids or is empty
-    IF (select array_agg(id) from collections) @> _collection_ids OR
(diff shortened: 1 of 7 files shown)


pypgstac/pypgstac/migrations/pgstac.0.6.9-0.6.10.sql
```diff
@@ -0,0 +1,42 @@
...
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION pgstac.get_queryables(_collection_ids text[] DEFAULT NULL::text[])
+ RETURNS jsonb
+ LANGUAGE plpgsql
+ STABLE
+AS $function$
+BEGIN
+    -- Build up queryables if the input contains valid collection ids or is empty
+    IF (select array_agg(id) from collections) @> _collection_ids OR
```
mmcfarland, line 11: Not sure if this is the right implementation.

bitner: Pushed a change that just switches this logic up with a bit more efficient approach using exists. Nice work with stringing together some PLPGSQL @mmcfarland !

mmcfarland: Nice, thanks!

Issue #152, opened 2023-01-03
Malformed query when using multiple sort fields and token


mmcfarland:
When using a token for paged search, the presence of multiple sortBy values _with different sort directions_ cause an incorrect SQL where clause to be generated.

In the `get_token_filter` function, all the provided sorts are inserted to the `sorts` temp table and the corresponding attributes from the Item represented by the token are meant to be used as the sort row's value. 

https://github.com/stac-utils/pgstac/blob/da8d6935a6b50b1eeaf226f38d02ca126cc62034/sql/004_search.sql#L333-L334

However, since the sort record is using the queryable.expression value, the `_field` value tend to look like, e.g. `to_int(content->'properties'->'eo:cloud_cover')`. I think only base attributes (`id`, `geometry`, etc) would work. I looked into using queryable.path as an alternative, but it's still a formatted string and I'm not sure how to transform a string representing a nested key to an actual object accessor. 

The end result is that `_val` is null for those records, and this results in where expressions like 

```sql
(("to_text(content->'properties'->'created')" < ) OR ("to_text(content->'properties'->'created')" =  AND
```

Here's a query that demonstrates the issue:

```sql
select search('{
	"datetime": "2022-11-20T00:00:00Z/2022-11-30T00:00:00Z",
	"collections": [
		"sentinel-2-l2a"
	],
	
	"sortby": [
		{
			"field": "properties.created",
			"direction": "desc"
		},
		{
			"field": "properties.eo:cloud_cover",
			"direction": "asc"
		}
	],
	"limit": 10,
	"token": "next:S2B_MSIL2A_20220727T191919_R099_T10UDV_20220729T035511"
}')
```

PR #153, opened 2023-01-09
Fix for Sorting and Upserts

bitner:
- Fix for issue with sorting (#152 )
- Fix for Upserts using items_staging_upsert table or upsert_item function.
- Add ability to merge enum, min, and max from queryables where collections have different values.
- Add tooling in pypgstac and pgstac to add stac_extension definitions to the database.
- Modify missing_queryables function to try to use stac_extension definitions to populate queryable definitions from the stac_extension schemas.
- Add validate_constraints procedure
- Add analyze_items procedure
- Add check_pgstac_functions function


bitner:
Hmm, I can't duplicate the black error locally that is failing in the tests (on a file that was not updated).

mmcfarland (approved):

PR #155, opened 2023-01-31
Maintenancetools

bitner:


src/pypgstac/pyproject.toml
```diff
@@ -0,0 +1,129 @@
...
+    "pytest",
+    "pytest-cov",
+    "pystac[validation]==1.*",
+]
+dev = [
+    "flake8==3.9.*",
+    "black>=21.7b0",
+    "mypy>=0.910",
+    "types-orjson==0.1.1",
+    "ruff==0.0.225",
```
vincentsarago, line 44: pre-commit ruff version is not the same https://github.com/stac-utils/pgstac/pull/155/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R19

Could we just use `pre-commit` directly without having to set a dev dependency?

bitner:
Reorganized PR for easier review in https://github.com/stac-utils/pgstac/pull/160

PR #160, opened 2023-02-06
Reorg0.7.0

### LINKED ISSUE #156: Make connection command-line parameters optional
It is not currently possible to connect to RDS Proxy using pypgstac due to its use of command-line parameters. Connection attempts fail with

> connection to server at "...rds.amazonaws.com" (IP), port 5432 failed: FATAL:  Feature not supported: RDS Proxy currently doesn’t support command-line options.

I've also seen a similar error when working with pgbouncer.

I had to override the `PgstacDB` class with the following to make this work

```python
class _ProxyCompatiblePgstacDB(PgstacDB):
    def get_pool(self) -> ConnectionPool:
        if self.pool is None:  # type: ignore
            self.pool = ConnectionPool(
                conninfo=self.dsn,
                min_size=pypgstac_settings.db_min_conn_size,
                max_size=pypgstac_settings.db_max_conn_size,
                max_waiting=pypgstac_settings.db_max_queries,
                max_idle=pypgstac_settings.db_max_idle,
                num_workers=pypgstac_settings.db_num_workers,
            )
        return self.pool

    def connect(self) -> Connection:
        connection = super().connect()
        connection.execute("set search_path = pgstac,public")
        return connection
```

Perhaps there can be a configuration option, or a variation on the above, to avoid always attempting to send command-line options.

### LINKED ISSUE #162: Allow use of pypgstac loader within same major version.
Currently pypgstac requires that the version of pypgstac exactly matches the version of pgstac it is being used against. When using the loader we should allow the use of pypgstac that is the same minor version ie "v0.6.1" pypgstac should be able to still load data into a "v0.6.5" database but not a "v0.7.0" database.

Any breaking changes to how the pypgstac loader works would need to happen at a major release.

bitner:
- Reorganize code base to create clearer separation between pgstac sql code and pypgstac.
- Move Python tooling to use hatch with all python project configuration in pyproject.toml
- Rework testing framework to not rely on pypgstac or migrations. This allows to run tests on any code updates without creating a version first. If a new version has been staged, the tests will still run through all incremental migrations to make sure they pass as well.
- Add pre-commit to run formatting as well as the tests appropriate for which files have changed.
- Add a query queue to allow for deferred processing of steps that do not change the ability to get results, but enhance performance. The query queue allows to use pg_cron or similar to run tasks that are placed in the queue.
- Modify triggers to allow the use of the query queue for building indexes, adding constraints that are used solely for constraint exclusion, and updating partition and collection spatial and temporal extents. The use of the queue is controlled by the new configuration parameter "use_queue" which can be set as the pgstac.use_queue GUC or by setting in the pgstac_settings table.
- Reorganize how partitions are created and updated to maintain more metadata about partition extents and better tie the constraints to the actual temporal extent of a partition.
- Add "partitions" view that shows stats about number of records, the partition range, constraint ranges, actual date range and spatial extent of each partition.
- Add ability to automatically update the extent object on a collection using the partition metadata via triggers. This is controlled by the new configuration parameter "update_collection_extent" which can be set as the pgstac.update_collection_extent GUC or by setting in the pgstac_settings table. This can be combined with "use_queue" to defer the processing.
- Add many new tests.
- Migrations now make sure that all objects in the pgstac schema are owned by the pgstac_admin role. Functions marked as "SECURITY DEFINER" have been moved to the lower level functions responsible for creating/altering partitions and adding records to the search/search_wheres tables. This should open the door for approaches to using Row Level Security.
- Set search_path and application_name after connection on pypgstac rather than as a kwarg parameter for compatibility with RDS (Fixes #156).
- Allow pypgstac loader to load data on pgstac databases that have the same major version even if minor version differs. Fixes #162. Cherry picked from #164 (Thanks @drnextgis).


src/pypgstac/pypgstac/load.py
```diff
@@ -1,27 +1,29 @@
...
+    TextIO,
     Tuple,
     Union,
-    Generator,
-    TextIO,
 )
+
 import orjson
 import psycopg
+from methodtools import lru_cache
```
vincentsarago, line 26: methodtools (which I didn't know) seems pretty old and not widely used, is there any reason not to use https://cachetools.readthedocs.io/en/latest/#module-cachetools.func ?

src/pypgstac/tests/hydration/test_dehydrate_pg.py
```diff
@@ -14,7 +16,6 @@ class TestDehydratePG(TDehydrate):
     @contextmanager
     def db(self) -> Generator:
         """Set up database connection."""
-        print("Setting up db.")
```
vincentsarago, line 17: 👋

vincentsarago (approved):

mmcfarland (approved):

src/pypgstac/tests/hydration/test_dehydrate_pg.py
bitner: Believe it or not, I don't think that one was me!

src/pypgstac/pypgstac/load.py
bitner: I didn't know there was an issue until Ruff said that the regular lru_cache has a memory issue and methodtools was the first thing googling came up with as a direct replacement. I'll change to cachetools.

src/pypgstac/tests/hydration/test_dehydrate_pg.py
vincentsarago: oh yeah I do print in testing too :D

src/pypgstac/pypgstac/db.py
```diff
@@ -86,7 +87,7 @@ def get_pool(self) -> ConnectionPool:
                 num_workers=settings.db_num_workers,
                 kwargs={
                     "options": "-c search_path=pgstac,public"
-                    " -c application_name=pypgstac"
+                    " -c application_name=pypgstac",
```
drnextgis, line 90: @bitner could you please make `PgstacDB` more robust by not using command-line options? As @captaincoordinates [pointed out](https://github.com/stac-utils/pgstac/issues/156) `PgstacDB`  cannot work properly with Amazon RDS Proxy in its current state. The possible solution would be to remove that, and then once the connection is established execute:
```
self.connection.execute(
    "set search_path = pgstac,public; set application_name=pypgstac"
)
```

bitner: @drnextgis @captaincoordinates, I've made a change for this in the latest commit. Can you verify that this works with RDS Proxy?

captaincoordinates: @bitner thanks for adding this change, unfortunately it will be a little while before I can test this on RDS Proxy - I'm currently using pypgstac via stac-fastapi and it doesn't support 0.7.0. I'll have to create a new project and some new AWS infrastructure for it to use before I can verify.

PR #164, opened 2023-02-14
Allow use of pypgstac loader within same minor version

### LINKED ISSUE #162: Allow use of pypgstac loader within same major version.

drnextgis:
Fixes #162

### DIFF
--- a/pypgstac/pypgstac/load.py
+++ b/pypgstac/pypgstac/load.py
@@ -24,6 +24,7 @@ import orjson
 import psycopg
 from orjson import JSONDecodeError
 from plpygis.geometry import Geometry
+from pkg_resources import parse_version as V
 from psycopg import sql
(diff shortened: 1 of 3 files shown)


drnextgis:
The failure is not related to the changes in this PR.

bitner:
@drnextgis Thanks for this patch! I wanted to get it into the PR that I'm staging to become the 0.7.0 release that has some reorganization of the pypgstac code, so I've cherry picked into PR #160

PR #175, opened 2023-04-18
Remove use of temp tables / cursor in search

bitner:
### Added
- Add --v and --vv options to scripts/test to change logging to notice / log when running tests.
- Add framework for option to cache expensive item formatting/hydrating calls. Note: this only provides functionality to add and read from the cached calls, but does not have any wiring to remove any entries from the cache.
- Update the costs for json formatting functions to 5000 to help the query planner choose to prefer using indexes on json fields.

### Fixed
- Fix bug in foreign key and unique collection detection in queryables trigger function, update tests to catch.
- Add collection id to tokens to ensure uniqueness and improve speed when looking up token values. Update tests to use the new keys. Old item id only tokens are still valid, but new results will all contain the new keys.
- Update Search function to remove the use of cursors and temp tables.
- Update get_token_filter to remove the use of temp tables.


src/pgstac/sql/004_search.sql
```diff
@@ -607,164 +603,283 @@ BEGIN
...
+CREATE INDEX ON format_item_cache (lastused);
+
+CREATE OR REPLACE FUNCTION format_item(_item items, _fields jsonb DEFAULT '{}', _hydrated bool DEFAULT TRUE) RETURNS jsonb AS $$
+DECLARE
+    cache bool := get_setting_bool('format_cache');
+    _output jsonb := null;
+    t timestamptz := clock_timestamp();
+BEGIN
+    IF cache THEN
+        SELECT output INTO _output FROM format_item_cache
```
mmcfarland, line 712: Is this a buyers-beware situation for folks opting into `format_cache`, any updates to the real item may not be reflected in search if the cache table wasn't also updated (deleted) on that id/collection?

mmcfarland (approved):

src/pgstac/sql/004_search.sql
bitner: There is a trigger on update/delete on items that will delete an item from the cache when the value has changed, but there is nothing in place to keep the cache from growing infinitely.

PR #179, opened 2023-04-28
Fix collection check in queryable trigger_func

mmcfarland:
The previous check for the existence of a collection could still return values
that weren't in the new collection_ids array.  This resulted in pgstac incorrectly
failing the check and preventing new queryables being added. This issue was only
present when there were more collections than just the collection being
tested. I added an additional test collection, added a test to verify
the error and then updated the sql join to only return results if there
is a missing collection.

Also makes a small adjustment to the queryable docs to use a more specific/correct insert query.


bitner (approved):
Looks good. Just needs a note added to the changelog.

PR #187, opened 2023-06-10
fix: return '[]' instead of NULL from all_collections when the table is empty

### LINKED ISSUE #186: all_collections returns null on an empty table
Running `all_collections` when there are no rows in the collections table returns null rather than an empty array. This breaks (imo) the implied guarantee that the function always returns valid JSON if it succeeds.

To repro, initialise an empty database by running `docker run -e POSTGRES_PASSWORD=password -p 5432:5432 ghcr.io/stac-utils/pgstac:v0.7.9`, then run `select all_collections();`:
```
postgres=# select all_collections();
 all_collections
-----------------

(1 row)
```

This is due to the behaviour of `jsonb_agg` here: https://github.com/stac-utils/pgstac/blob/main/src/pgstac/sql/002_collections.sql#L66

Elsewhere pgstac uses `coalesce` to convert nulls to `[]`: https://github.com/stac-utils/pgstac/blob/main/src/pgstac/sql/006_tilesearch.sql#L103, I think the same thing should be done for `all_collections()`.

mrob95:
fixes #186 

test fails before:
image

test passes after:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Fixed
+- Return an empty jsonb array from all_collections() when the collections table is empty, instead of NULL. Fixes #186.
+
 ## [v0.7.9]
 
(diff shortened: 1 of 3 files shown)


bitner (approved):
Thanks for catching this. I'll merge this and then get a release out with this this week.

PR #199, opened 2023-08-15
Rusthydrate

bitner:


.gitignore
```diff
@@ -1,7 +1,9 @@
+.so
```
kylebarron, line 1: Actually maybe this needs `*.so` to match the `.so` file generated

bitner:
Refactored into PR #201

PR #197, opened 2023-07-31
udpate pydantic to 2.0

vincentsarago:
This PR does:
- update pydantic requirement from `==1.10.*` to `~=2.0`: Looking at the usage of pydantic in this project, I don't think we have to be really narrow in the `pin` and using any `2.*` version should be good. 

FYI: This is blocking titiler-pgstac update

PR #201, opened 2023-08-21
Rust Hydration

bitner:
- update `pydantic` requirement to `~=2.0`
- update docker and ci workflows to build binary wheels for rust additions to pypgstac
- split docker into database service and python/rust container
- Modify scripts to auto-generate unreleased migration
- Add pre commit tasks to generate migration and to rebuild and compile pypgstac with maturin for rust
- Add private jsonb column to items and collections table to hold private metadata that should not be returned as part of a stac item
- Add generated columns to collections with the bounding box as a geometry and the datetime and end_datetime from the extents (this is to help with forthcoming work on collections search)
- Add PLRust to the Docker postgres image for forthcoming work to add optional PLRust functions for expensive json manipulation (including hydration)


src/pypgstac/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+path = "src/lib.rs"
+name = "pgstacrs"
+crate-type = ["cdylib"]
+
+[dependencies]
+pyo3 = { version = "0.19.1", features = ["abi3-py38", "extension-module"] }
```
kylebarron, line 13: Note that pyo3 also has a few more features that might be useful to you. In particular `serde`, `anyhow`, `chrono`, and `hashbrown` could be useful. Especially since you're already using `serde` and `thiserror` separately (I think `thiserror` integrates well with anyhow IIRC)

src/pypgstac/pyproject.toml
```diff
@@ -38,6 +39,9 @@ test = [
     "pystac[validation]==1.*",
     "types-cachetools",
 ]
+build = [
+    "maturin>=1,<2",
+]
```
kylebarron, line 44: You can define this if you want, but it's superfluous with `build-system` below. If you use any PEP-517 compliant build tool, `maturin` will be installed automatically. E.g.

```bash
pip install -U build
python -m build .
```
will install maturin into the build environment automatically.

src/pypgstac/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+struct HydrateError(Error);
+
+impl From<HydrateError> for PyErr {
+    fn from(value: HydrateError) -> Self {
+        HydrationError::new_err(value.0.to_string())
+    }
+}
```
kylebarron, line 70: Why do you have `HydrateError` as a separate type than `Error`? Seems like it would be straightforward to have only a single type and implement `PyErr` on that type?

src/pypgstac/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+#[pyfunction]
+fn hydrate(base: &PyAny, item: &PyAny) -> PyResult<Py<PyAny>> {
+    let mut serde_item: Value = pythonize::depythonize(item)?;
+    let serde_base: Value = pythonize::depythonize(base)?;
+    serde_item.hydrate(serde_base).map_err(HydrateError)?;
+    pythonize::pythonize(item.py(), &serde_item).map_err(PyErr::from)
```
kylebarron, line 76: How does `pythonize` differ from the standard serde integration in pyo3? Is it supposed to be faster?

kylebarron (approved):
I'm not very familiar with the postgres internals of pgstac, but the Rust code and Rust CI changes look good

src/pypgstac/src/lib.rs
gadomski, line 1: @bitner have you done any profiling to ensure that this is faster than a pure Python implementation? I did some back-of-envelope benchmarking of a couple of different Python<->Rust implementations, and IIRC this way (going all the way to a `serde_json::Value` was significantly slower than both a pure Python and a "use Rust to manipulate a Python dictionary" options. I can't find that code at the moment, but if I have some spare time I'll re-run the experiment to be sure.

gadomski: I did some quick benchmarking here: https://github.com/gadomski/json-hydrate-benchmark. Tl;dr: going all the way to a `serde_json::Value` is slower than just doing it in Python, **but** modifying the Python objects directly w/ the pyo3 interface is faster!

| method | time |
| -- | -- |
| python | 43.2±2μs |
| serde_json | 193±10μs |
| pyo3 | 7.89±0.2μs |


[Here's what that looks like (error handling is ad-hoc)](https://github.com/gadomski/json-hydrate-benchmark/blob/712fad9277fbcb148f949eb4e951c3319f2fa380/src/lib.rs#L67-L120):

```rust
#[pyfunction]
fn pyo3<'a>(base: &PyAny, item: &'a PyAny) -> PyResult<&'a PyAny> {
    fn hydrate<'a>(base: &PyAny, item: &'a PyAny) -> PyResult<&'a PyAny> {
        if let Ok(item) = item.downcast::<PyDict>() {
            if let Ok(base) = base.downcast::<PyDict>() {
                hydrate_dict(base, item).map(|item| item.into())
            } else {
                Err(anyhow!("type mismatch").into())
            }
        } else if let Ok(item) = item.downcast::<PyList>() {
            if let Ok(base) = base.downcast::<PyList>() {
                hydrate_list(base, item).map(|item| item.into())
            } else {
                Err(anyhow!("type mismatch").into())
            }
        } else {
            Ok(item)
        }
    }


    fn hydrate_list<'a>(base: &PyList, item: &'a PyList) -> PyResult<&'a PyList> {
        for i in 0..item.len() {
            if i >= base.len() {
                return Ok(item.into());
            } else {
                item.set_item(i, hydrate(&base[i], &item[i])?)?;
            }
        }
        Ok(item)
    }


    fn hydrate_dict<'a>(base: &PyDict, item: &'a PyDict) -> PyResult<&'a PyDict> {
        for (key, base_value) in base {
            if let Some(item_value) = item.get_item(key) {
                if item_value
                    .downcast::<PyString>()
                    .ok()
                    .and_then(|value| value.to_str().ok())
                    .map(|s| s == MAGIC_MARKER)
                    .unwrap_or(false)
                {
                    item.del_item(&key)?;
                } else {
                    item.set_item(key, hydrate(base_value, item_value)?)?;
                }
            } else {
                item.set_item(key, base_value)?;
            }
        }
        Ok(item)
    }


    hydrate(base, item)
}
```

kylebarron: > I did some quick benchmarking here: https://github.com/gadomski/json-hydrate-benchmark. Tl;dr: going all the way to a `serde_json::Value` is slower than just doing it in Python, **but** modifying the Python objects directly w/ the pyo3 interface is faster!

Cool! I suppose the only downside of using the pyo3 interface directly is that you're holding the GIL the entire time. Which in this case is probably fine anyways unless you really wanted to try and multithread this.

kylebarron: It looks like it may be getting faster in the future too! https://slides.com/samuelcolvin/deck-a16632#/13

vincentsarago:
@bitner, This PR is blocking some other PR and projects. What do you think about splitting this into multiple PR (e.g https://github.com/stac-utils/pgstac/pull/197) so we can move forward.

bitner:
Main blocker is me being out of the office. I'm hoping to get everything merged on Monday. If i hit any road blocks I'll split things up

bitner:
@kylebarron @gadomski I updated to use pyo3 rather than serde using sample code from @gadomski. As we discussed last week, my main goal here is to get the infrastructure into place for incorporating rust with hopefully a bit of a win on hydration performance, but more with an eye towards moving much of the json parsing as well as things like injecting links into rust in the future.

scripts/pgstacenv
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+if [[ "${CI}" ]]; then
+    set -x
+fi
+
+function usage() {
+    echo -n \
+        "Usage: $(basename "$0")
+Format code in this project
```
gadomski, line 16: Looks like an invalid usage text, doesn't appear to be a format script.

src/pypgstac/Cargo.lock
gadomski, line 1: This file shouldn't be checked in, as it's a library (not an executable) (unless pyo3 projects are different cc @kylebarron).

src/pypgstac/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pgstacrs"
+version = "0.7.10-dev"
```
gadomski, line 3:
```suggestion
version = "0.0.0"
```

Pre-releases aren't really supported: https://github.com/rust-lang/cargo/issues/2222. Also, I don't think we'll be releasing this crate on its own -- it's only meant for use in the Python bindings. So `0.0.0` will indicate that it's not meant for release to crates.io.

src/pypgstac/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pgstacrs"
```
gadomski, line 2:
```suggestion
name = "pypgstac"
```

We already have a **pgstac**: https://crates.io/crates/pgstac. The `rs` is redundant, since this is a Rust package definition. IMO `pypgstac` would be the best Rust package name.

src/pypgstac/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+path = "src/lib.rs"
+name = "pgstacrs"
+crate-type = ["cdylib"]
+
+[dependencies]
+anyhow = "1"
+pyo3 = { version = "0.19.1", features = ["abi3-py38", "extension-module", "anyhow"] }
+pythonize = "0.19"
```
gadomski, line 15:
```suggestion
```

Don't need **pythonize** anymore if we're going **pyo3**-native route.

gadomski (changes requested):

gadomski (approved):

.gitignore
```diff
@@ -10,3 +10,4 @@ src/pypgstac/target
 src/pypgstac/python/pypgstac/*.so
 .vscode
 .ipynb_checkpoints
+Cargo.lock
```
kylebarron, line 13: This is up to you; in my projects I include the lockfile even for libraries so that developers at the same commit or checking out the same PR are always using the exact same dependency versions 🤷‍♂️

src/pypgstac/Cargo.toml
```diff
@@ -1,17 +1,14 @@
...
 name = "pgstacrs"
-version = "0.7.10-dev"
+version = "0.0.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [lib]
 path = "src/lib.rs"
-name = "pgstacrs"
+name = "pypgstac"
```
kylebarron, line 9: I think this is the name of the python package generated by the maturin. I'd recommend for it to be something like `_rust` or `_pgstacrs` so that 1) it's clear it's internal code not expected to be called by the user (re-exporting anything that we need directly through python code) and 2) making it clear when reading the code when a function is imported from the generated rust.

As it is, `from pypgstac.pypgstac import *` would I think import Rust functions, but it's easy to forget that when reading

src/pypgstac/src/lib.rs
```diff
@@ -1,83 +1,68 @@
...
-                _ => Err(Error::TypeMismatch(self.clone(), base)),
-            },
-            Value::Array(item) => match base {
-                Value::Array(base) => item.hydrate(base),
-                _ => Err(Error::TypeMismatch(self.clone(), base)),
-            },
-            _ => Ok(()),
+#[pyfunction]
+fn hydrate<'a>(base: &PyAny, item: &'a PyAny) -> PyResult<&'a PyAny> {
+    fn hydrate_any<'a>(base: &PyAny, item: &'a PyAny) -> PyResult<&'a PyAny> {
```
kylebarron, line 11: I wouldn't have usually written inner functions in the internal scope of the python function... I wonder if that's in effect just stylistic or whether there's any compilation performance that changes 🤷‍♂️

src/pypgstac/Cargo.toml
kylebarron: Also since the `#[pymodule]` is still set to `pgstacrs`, not sure if this builds correctly?

gadomski: Yeah, you'd have to change the name over there too, and probably in the python code.

.gitignore
gadomski: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html recommends ignoring it for libraries. That same page recommends version-controlling `Cargo.lock` for `cdylib`, so we should keep it.

src/pypgstac/Cargo.toml
kylebarron: It unexpectedly seems to still build for me... I can't tell if it's actually working or if maybe I'm loading cached artifacts

.gitignore
kylebarron: Interesting; I guess `cdylib` making a system library is in effect an application!

kylebarron (approved):

gadomski (approved):
@bitner I made a couple of touchups to the Rust code and the pypgstac Dockerfile. LGTM for a first-pass, :shipit:.

bitner:
Thanks @gadomski and @kylebarron for all the review / code! I'm going to merge this now, but not going to cut a release until I do some heavy benchmarking with PC.

PR #204, opened 2023-08-24
Updating base image to debian bookworm, and generating an image with postgres 15.4

hectormachin:
Resolves [CVE-2023-39417](https://www.postgresql.org/support/security/CVE-2023-39417/) and several other HIGH & CRITICAL CVEs have been reported with postgres 15.3 and older versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM postgres:15-bullseye as pg
+FROM postgres:15-bookworm as pg
 ENV PGSTACDOCKER=1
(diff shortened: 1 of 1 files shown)


bitner:
The latest push to main removed the pin on specific point version of postgres and postgis which updates postgres to 15.4 and postgis to 3.4. Is there a specific reason that you changed from bullseye to bookworm? If not, can we consider this ticket closed?

As soon as I can do some benchmarking on the main branch (this week), I will be pushing that out as a v0.8.0 release.

PR #208, opened 2023-09-03
Revert id search limit shortcut - #192

### LINKED ISSUE #192: Search does not account for items that are in multiple collections
According to the spec, item ids "should be unique within the Collection that contains the Item." (https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields).

This means that it is possible to have an item with the same ID in multiple collections. Pgstac does not take this into account when constructing search queries though. For example a query with body:
```
{
    "ids": [
        "fc92ee93-5707-4110-91e9-c4ed000a7462"
    ],
    "collections": [
        "b6eb0ede-9eed-4e7e-aa9c-b50887ce252d",
        "4f939cb1-58cf-470a-b1ba-274c641068c4"
    ]
}
```

Runs the following SQL (version 0.7 .10):
```
2023-07-03 16:06:32.099 UTC [924] STATEMENT:  SELECT search($1::jsonb);
2023-07-03 16:06:32.110 UTC [924] LOG:  QUERY:
           SELECT * FROM items
           WHERE
           datetime >= '-infinity' AND datetime < 'infinity'
           AND (id = ANY ('{fc92ee93-5707-4110-91e9-c4ed000a7462}')  AND collection = ANY ('{b6eb0ede-9eed-4e7e-aa9c-b50887ce252d,4f939cb1-58cf-470a-b1ba-274c641068c4}') )
           ORDER BY datetime DESC, id DESC
           LIMIT '1'
```

The `LIMIT '1'` is incorrect here as there could be an item with that ID in both of the collections passed.

mrob95:
Fixes #192 

I thought about a couple of other options, like limiting the optimisation to only kick in when there is only one collection specified in the query, or making it `items_cnt*collections_cnt`. I think both of these could be better implemented as a uniqueness constraint on `(id, collection)` in the items table though - postgres *should* be smart enough to figure out how many items to expect in that case.


mrob95:
The tests seem to fail unless I run `scripts/stageversion 0.7.11` first, so I've gone ahead and committed this

bitner:
The uniqueness of id, collection is *mostly* ensured by the partition constraint on collection and the unique constraint within each partition on id. From a constraint standpoint, it is technically possible for the same id to exist within the same collection if they come in with datetimes that put them into different partitions. It is not possible for unique constraints to extend beyond a partition in Postgres. Adding a unique constraint on (id, collection) is not possible on the items parent table and is redundant and counterproductive when placed on the partitions.

I think that the best we can do is to limit to items_cnt*collections_cnt. I'll make that change on top of the latest changes in the main branch and then add the tests from this PR. 

Thanks @mrob95 !

bitner:
@mrob95 due to conflicts with intermediate commits, I cherry picked these onto current main into PR #212 . The new build system now allows for unreleased versions which I hope should make PRs like this easier in the future.

PR #212, opened 2023-09-20
Revert id search limit shortcut - #192

### LINKED ISSUE #192: Search does not account for items that are in multiple collections

bitner:
Fixes #192 

Cherry Picked from #208 due to conflicts with intermediate commits, thank you @mrob95!


mrob95:
Many thanks!

PR #215, opened 2023-09-25
Read only configuration mode

### LINKED ISSUE #191: STAC Server responding with SQL Create Error when using /collections/<id>/items with dedicated STAC Reader
In our setup with a dedicated AWS Aurora READ replica as Reader, the STAC server responds with this error message when querying the collections/_collection_id/items endpoints.
{
    "code": "ReadOnlySQLTransactionError",
    "description": "cannot execute CREATE TABLE in a read-only transaction"
}

The issues obviously disappears when setting the reader to the read/write instance however, I assume having a writeable instance as READER  wouldn't be the intention.

bitner:
Fixes #191

This adds a configuration parameter "readonly".

The "readonly" configuration mode prevents searches and context functions to store data in the _searches and _search_wheres table. Setting "readonly" prevents the lookup of searches using a search hash (as used in TiTiler-PGStac) and does not cache context lookups, so use of this with context turned on will be slow for paged results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [unreleased]
+- Adds read only mode to allow use of pgstac on read replicas
+ - Note: Turning on romode disables any caching (particularly when context is turned on) and does not allow to store q query hash that can be used with geometry_search.
+
 ## [v0.8.1]
 
 ### Fixed
--- a/docs/src/pgstac.md
+++ b/docs/src/pgstac.md
@@ -113,6 +113,8 @@ The `queryables` table controls the indexes that PGStac will build as well as th
 
 Each record in the queryables table references a single property but can apply to any number of collections. If the `collection_ids` field is left as NULL, then that queryable will apply to all collections. There are constraints that allow only a single queryable record to be active per collection. If there is a queryable already set for a property field with collection_ids set to NULL, you will not be able to create a separate queryable entry that applies to that property with a specific collection as pgstac would not then be able to determine which queryable entry to use.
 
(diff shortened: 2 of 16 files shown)


mmcfarland (approved):
Looks good. There are some important consequences to enabling this mode, consider adding a section to the readme/docs that explain its existence, why you might want it enabled, and what you can't do.  As it relates to read replicas, the linked issue mentions that the stac-fastapi-pgstac implementation uses the read host for executing searches (which should use the write host). If that change was suitable to support read-replicas, what's the use case for a system setting that covers all connections?

I believe that `titiler-pgstac` just executes a search to "register" it, so that's why that function will no longer be possible. In future work, would it be possible for a specific write operation to hash and save to `searches` so that one could still register a search but still do a read-only search operation when consuming it?

bitner:
> Looks good. There are some important consequences to enabling this mode, consider adding a section to the readme/docs that explain its existence, why you might want it enabled, and what you can't do. 

As it relates to read replicas, the linked issue mentions that the stac-fastapi-pgstac implementation uses the read host for executing searches (which should use the write host). If that change was suitable to support read-replicas, what's the use case for a system setting that covers all connections?

I think that perhaps we have a bit of a misnomer that we are using in stac-fastapi-pgstac. When I think read vs write in the context of "read host" and "write host", I am thinking about the actual records themselves, so the "write host" is used with the transactions api and the "read host" is used for everything else. I've never really counted the statistics, cache, or registering of a search that get written to the database as "writes" from a stac perspective. It might make sense to make some changes in stac-fastapi-pgstac that allow for a flag to be used that indicates whether a search should go to the "read host" (which could be a read replica *if* the readonly setting was turned on) or to the configured "write host" or create yet a third host and make the setting names a bit more clear. This flag could then be set by titiler-pgstac only when registering a search.

> 
> I believe that `titiler-pgstac` just executes a search to "register" it, so that's why that function will no longer be possible. In future work, would it be possible for a specific write operation to hash and save to `searches` so that one could still register a search but still do a read-only search operation when consuming it?

Yes, it is only the registering of a search that would not be possible. It would be possible to direct the register only to the primary database and then all titiler-pgstac calls that use that hash to the read replica.

PR #219, opened 2023-10-04
Make sure that only arguments that are properties are cheked

### LINKED ISSUE #218: Exception with a legitimate SQL query
Disabling `additionalProperties` results in the following query triggering an exception::
```
postgis=# SELECT search('{"filter": {"and": [{"t_after": [{"property": "datetime"}, "2020-11-11T00:00:00"]}, {"t_before": [{"property": "datetime"}, "2022-11-11T00:00:00"]}]}}');
ERROR:  Term <NULL> is not found in queryables.
CONTEXT:  PL/pgSQL function cql2_query(jsonb,text) line 110 at RAISE
PL/pgSQL function stac_search_to_where(jsonb) line 60 at assignment
PL/pgSQL function search_query(jsonb,boolean,jsonb) line 18 at assignment
PL/pgSQL function search(jsonb) line 39 at assignment
```

drnextgis:
Fixes #218

### DIFF
--- a/src/pgstac/migrations/pgstac.0.8.1-unreleased.sql
+++ b/src/pgstac/migrations/pgstac.0.8.1-unreleased.sql
@@ -297,7 +297,7 @@ BEGIN
             END LOOP;
 
             IF
-                NOT extra_props AND wrapper IS NULL
+                NOT extra_props AND wrapper IS NULL AND jsonb_path_exists(args, '$[*] ? (@.property != null)')
             THEN
                 RAISE EXCEPTION 'Term % is not found in queryables.', arg->>'property';
             END IF;
--- a/src/pgstac/migrations/pgstac.0.8.1-unreleased.sql
+++ b/src/pgstac/migrations/pgstac.0.8.1-unreleased.sql
@@ -201,12 +201,27 @@ DECLARE
     _wrapper text;
     leftarg text;
     rightarg text;
+    prop text;
     extra_props bool := pgstac.additional_properties();
 BEGIN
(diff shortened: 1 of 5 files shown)


drnextgis:
I moved the code to a different location and added a loop to iterate through all properties, ensuring none of them are included in the queryables. Previously, the code was placed incorrectly in the loop which was only meant to detect the wrapper. Furthermore, I made it explicit in the code that the properties id, datetime, end_datetime, and collection are always considered queryable.

bitner (dismissed):
Good Catch!

bitner:
@drnextgis I think if we put something like this:
```
    IF NOT extra_props THEN
        FOR prop IN 
            SELECT DISTINCT p->>0 
            FROM jsonb_path_query(j, 'STRICT $..property')
            WHERE p->>0 NOT IN ('id', 'datetime', 'end_datetime', 'collection')
        LOOP
            IF (queryable(prop)).nulled_wrapper IS NULL THEN
                RAISE EXCEPTION 'Term % is not found in queryables.', arg->>'property';
            END IF;
        END LOOP;
    END IF;
```
at the very top of the cql2_query (probably right after the `RAISE NOTICE 'CQL2_QUERY: %', j;` it should find any property nested anywhere in the filter on the first call to cql2_query rather than when things have gone down the nesting iterations. I think this will be more resistant to any of the different implementations for spatial, array, and temporal operators.

drnextgis:
@bitner thank you! I've updated the PR.

bitner (approved):
Looks good!

PR #235, opened 2024-01-24
Fix for CI docker builds.

bitner:
The docker image for pgstac has four targets.

- pgstacbase: This target is essentially the same as postgis/postgis with pgtap, plpgsql-check, and partman installed (which are used for testing/development, but not required for using pgstac), it is a base postgres/postgis image that is ready to install pgstac, but does not have it installed yet.
- pgstacbase-plrust: This is the pgstacbase image that also has PLRUST installed in preparation for future development (PLRUST is not used yet).
- pgstac: This target is the pgstacbase image with pgstac already installed at a specific version
- pgstac-rust: This target is the pgstacbase-plrust image with pgstac already installed at a specific version

CI ran into issues with building the pgstacbase-plrust and pgstac-plrust targets. This PR does three things:

- Update the pgstacbase-plrust target with the latest PLRUST version
- Modify the CI release to push the pgstacbase and pgstac targets rather than the PLRUST targets as PLRUST is not used yet.
- Update the incremental migration tests to start at v0.3.0 rather than v0.1.9 due to a breaking change in pg_partman at version 5 that has no ability to pin a version.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,7 +47,7 @@ jobs:
 
   # This builds a base postgres image that has everything installed to be able to run pgstac. This image does not have pgstac itself installed.
   buildpg:
-    name: Build and push base postgres image (you will need to run pypgstac migrate to install pgstac)
+    name: Build and push base postgres image
     if: ${{ needs.changes.outputs.buildpgdocker == 'true' }}
     runs-on: ubuntu-latest
     needs: [changes]
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -45,8 +45,9 @@ jobs:
         echo "pytag=${{ env.REGISTRY }}/stac-utils/pgstac-pyrust:$ref" >>$GITHUB_OUTPUT;
         echo "buildpy=$buildpg" >>$GITHUB_OUTPUT;
 
+  # This builds a base postgres image that has everything installed to be able to run pgstac. This image does not have pgstac itself installed.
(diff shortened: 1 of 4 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -45,8 +45,9 @@ jobs:
         echo "pytag=${{ env.REGISTRY }}/stac-utils/pgstac-pyrust:$ref" >>$GITHUB_OUTPUT;
         echo "buildpy=$buildpg" >>$GITHUB_OUTPUT;
 
+  # This builds a base postgres image that has everything installed to be able to run pgstac. This image does not have pgstac itself installed.
   buildpg:
-    name: Build and push base postgres
+    name: Build and push base postgres image (you will need to run pypgstac migrate to install pgstac)
```
vincentsarago, line 50:
```suggestion
    name: Build and push base postgres image
```

I don't think we need this, It's fine with the comment 👍

.github/workflows/release.yml
```diff
@@ -21,9 +21,9 @@ jobs:
           repo_token: "${{ secrets.GITHUB_TOKEN }}"
           prerelease: false
 
-
+  # This builds a base postgres image that has everything installed to be able to run pgstac.
   buildpg:
-    name: Build and push base postgres
+    name: Build and push base postgres image (you will need to run pypgstac migrate to install pgstac)
```
vincentsarago, line 26:
```suggestion
    name: Build and push base postgres image
```

Same

vincentsarago (approved):

Issue #249, opened 2024-03-12
BETWEEN operator bug


drnextgis:
As per the [spec](https://github.com/opengeospatial/ogcapi-features/blob/7759bc88b36e56a2dbe0a254d92f119c9ce25fac/cql2/standard/schema/cql2.json#L134) and [examples](https://github.com/stac-api-extensions/filter/?tab=readme-ov-file#example-11-cql2-json-post), the `args` parameter for the `BETWEEN` operator should consist of a list containing three elements. However, it appears that PgSTAC anticipates receiving a list with only two elements, as inferred from [here](https://github.com/stac-utils/pgstac/blob/1b678bd7ea3f3e89b9bf947453a2cd3d2f476282/src/pgstac/sql/002b_cql.sql#L344).

PR #251, opened 2024-03-13
Fix CQL2 parser for BETWEEN operator

drnextgis:
Fixes https://github.com/stac-utils/pgstac/issues/249


drnextgis:
@bitner since there was already a test, I simply made a small update to it. Would that be sufficient?

bitner (approved):

PR #254, opened 2024-04-09
Add numberReturned and numberMatched fields to ItemCollection

drnextgis:
ref: https://github.com/radiantearth/stac-api-spec/issues/396
ref: https://github.com/stac-utils/stac-fastapi/issues/478

The context extension is deprecated.


drnextgis:
@bitner please review and let me know whether you believe we should also consider updating `collection_search`?

vincentsarago:
thanks @drnextgis 🙏

vincentsarago (approved):

bitner:
@drnextgis This looks great, if you are up to changing in collection_search as well, that would be awesome. I don't have strong feelings about this, but do we want an option to use the deprecated context still for any clients that continue to expect that?

drnextgis:
`collection_search` has been updated, and tests have been adjusted

drnextgis:
> @drnextgis This looks great, if you are up to changing in collection_search as well, that would be awesome. I don't have strong feelings about this, but do we want an option to use the deprecated context still for any clients that continue to expect that?

I agree, I don't have strong feelings either. Documenting it as a backwards incompatible change in the changelog should suffice. In stac-pydantic it was [removed](https://github.com/stac-utils/stac-pydantic/pull/138) by @vincentsarago.

vincentsarago:
> I don't have strong feelings either. Documenting it as a backwards incompatible change in the changelog should suffice. In stac-pydantic it was https://github.com/stac-utils/stac-pydantic/pull/138 by @vincentsarago.

The ContextExtension is deprecated and will be removed in next version of stac-fastapi https://github.com/stac-utils/stac-fastapi/issues/649

bitner (approved):

PR #256, opened 2024-04-16
Add support for casei and accenti

### LINKED ISSUE #237: CQL2: ILIKE vs. CASEI
CQL2 has evolved and the ILIKE operator was removed from the specification. pgstac still implements ILIKE. Instead there's now a function called CASEI for a similar purpose. (ACCENTI might also be relevant.) Those are not yet implemented. This should probably be made specification compliant at some point.

Spec: https://github.com/opengeospatial/ogcapi-features/blob/master/cql2/standard/annex_ats_case-insensitive-comparison.adoc

Related issues:
- https://github.com/opengeospatial/ogcapi-features/issues/630
- https://github.com/opengeospatial/ogcapi-features/issues/828

bitner:
Adds support for casei and accenti operators. Fixes #237. Does not remove existing upper, lower functions or the ilike operator to not break backwards compatibility.

### DIFF
--- a/src/pgstac/migrations/pgstac.0.8.5-unreleased.sql
+++ b/src/pgstac/migrations/pgstac.0.8.5-unreleased.sql
@@ -1,0 +1,746 @@
+SET client_min_messages TO WARNING;
+SET SEARCH_PATH to pgstac, public;
+RESET ROLE;
+DO $$
+DECLARE
+BEGIN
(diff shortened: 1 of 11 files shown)


m-mohr (approved):
I don't understand everything, but it generally looks good to me. Thanks for doing it!

PR #266, opened 2024-04-29
Paging as links

bitner:
Use standard compliant links for prev/next.


drnextgis:
I'm curious why new functions are created in the pg_temp schema instead of pgstac.

bitner:
> I'm curious why new functions are created in the pg_temp schema instead of pgstac.

@drnextgis Those are functions that are only used for the tests, so there is really no need to keep them around.

zacharyDez (approved):

PR #295, opened 2024-08-13
add q parameter for free-text search

### LINKED ISSUE #293: Support for the STAC API free-text extension
The [STAC API free-text extension](https://github.com/stac-api-extensions/freetext-search?tab=readme-ov-file) enables a user to provide a parameter `q` with some free-text search query parameters. This is perhaps redundant with the filter extension, but it is much easier to write a text query using the `q` parameter. e.g. `landsat OR sentinel` to get records that contain the term `sentinel` or `landsat`, `landsat AND sentinel` to get records that contain both terms.

Adding support for the `q` parameter would be particularly useful for improving the collection search functionality!

hrodmn:
This adds the `q` parameter for free-text search. It will be pretty slow for item-level search because it has to compute the `tsvectors` on-the-fly, but it works great for collection-level search where the number of records should be limited.

Resolves #293 (for collection-level search at least)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,43 @@
 # Changelog
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-### [v0.9.1]
+## Unreleased
+
+- Add limited support for free-text search in the search functions. (Fixes #293)
+  - the `q` parameter is converted from the
+    [OGC API - Features syntax](https://docs.ogc.org/DRAFTS/24-031.html) into a `tsquery`
+    statement which is used to compare to the description, title, and keywords fields in items or collection_search
+  - the text search is un-indexed and will be very slow for item-level searches!
+
+## [v0.9.1]
 
 ### Fixed
+
 - Fixed double nested extent when using trigger based update collection extent. (Fixes #274)
 - Fix time formatting (Fixes #275)
 - Relaxes smart-open dependency check (Fixes #273)
 - Switch to uv for docker image
 
-
-### [v0.9.0]
+## [v0.9.0]
 
 ### Breaking Changes
+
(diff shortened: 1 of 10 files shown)


src/pypgstac/pyproject.toml
```diff
@@ -46,7 +46,7 @@ dev = [
     "black>=21.7b0",
     "mypy>=0.910",
     "types-orjson==0.1.1",
-    "types-pkg-resources",
+    "types-setuptools",
```
hrodmn, line 49: I got an error when I built the docker containers saying `types-pkg-resources` had been "pulled" and to replace it with `types-setuptools`

src/pgstac/sql/004_search.sql
```diff
@@ -127,6 +127,32 @@ CREATE OR REPLACE FUNCTION partition_query_view(
...
+    RETURNS tsquery
+    AS $$
+DECLARE
+    processed_text text;
+BEGIN
+    -- replace commas outside quoted text with " | "
+    processed_text := regexp_replace(input, ',(?=(?:[^"]*"[^"]*")*[^"]*$)', ' | ', 'g');
+
+    -- replace the logical operators with tsquery equivalents
+    processed_text := regexp_replace(processed_text, '\s+AND\s+', ' & ', 'g');
```
bitner, line 140: can we make this case insensitive?

CHANGELOG.md
hrodmn, line 1: I think my editor automatically applied some formatting changes to this file. I can roll them back if you want.

src/pgstac/sql/004_search.sql
```diff
@@ -139,8 +165,10 @@ DECLARE
     sdate timestamptz;
     edate timestamptz;
     filterlang text;
-    filter jsonb := j->'filter';
+    filter jsonb = j->'filter';
```
bitner, line 168: this should stay ":=" for assignment in plpgsql

hrodmn:
@bitner I had to beef up `q_to_tsquery` a bit to avoid replacing terms inside of quoted statements. It's all case-insensitive now and passing a more robust set of tests!

bitner (approved):

Issue #296, opened 2024-08-19
Error: Cannot execute INSERT in a read-only transaction


jose-lpa:
Hello. First please apologise if this has been asked before. I am having a blocker issue with PgSTAC when trying to run it in AWS cloud using a Postgres RDS setup.

The RDS setup consists of a PostgreSQL cluster with a **master** instance for **writing** data and a **replica** for **reading** it.

In this standard setup, the replica instance is automatically set up as **read-only**. This effectively means that **no `INSERT` transactions are accepted by this PostgreSQL instance**. Unfortunately, this doesn't play well at all with [the logic behind the STAC search functionality](https://github.com/stac-utils/pgstac/blob/febbaf8f0b085968e04e6d2b2770bfc403f2ff79/src/pgstac/sql/004_search.sql). As I can see in PgSTAC implememntation, there are quite a lot of `INSERT` operations during a regular STAC search. This makes impossible to have the cloud database setup I was thinking on: master-replica(s), where the master is only used for writing and all the replicas for reading, e.g. "searching" or "consulting" items.

This master-replica implementation is not uncommon or something customised by me. Instead, it kind of comes as default setup in cloud providers such as AWS and GCP when you aim to start scaling the system (I've checked in both, and both have the same problem).

My question here is: where are you guys running PgSTAC? How do you configure a high-availability setup with replicas, if I may ask? Where can I find documentation about deploying PgSTAC in a cloud environment?

Many thanks for your time.

drnextgis:
@jose-lpa please take a look at this https://github.com/stac-utils/pgstac/pull/215

drnextgis:
I experienced a similar issue, though with a slight difference:

```sql
postgis=> SELECT get_version();
 get_version 
-------------
 0.9.1
(1 row)

postgis=> SELECT pgstac.readonly();
 readonly 
----------
 t
(1 row)

postgis=> SHOW transaction_read_only;
 transaction_read_only 
-----------------------
 on
(1 row)

postgis=> SELECT search('{}');
NOTICE:  SEARCH: {}
NOTICE:  FILTER: <NULL>
ERROR:  cannot execute SELECT FOR UPDATE in a read-only transaction
CONTEXT:  SQL statement "SELECT *         FROM search_wheres WHERE md5(_where)=inwhere_hash FOR UPDATE"
PL/pgSQL function where_stats(text,boolean,jsonb) line 57 at SQL statement
PL/pgSQL function search(jsonb) line 65 at assignment
```

[...]

PR #297, opened 2024-08-20
Add test to verify search works in readonly mode with context extensi…

drnextgis:
…on enabled


drnextgis:
The test is expected to pass, but it fails, reproducing the issue I described [here](https://github.com/stac-utils/pgstac/issues/296#issuecomment-2297681742). This appears to be a bug, and the code may need adjustment. Although the `context` extension is enabled in the write replica (via the `pgstac_settings` table), it shouldn't impact the read-only replica.

bitner (approved):

PR #304, opened 2024-09-17
switch from server_version to server_version_num

vincentsarago:
ref: https://github.com/stac-utils/stac-fastapi-pgstac/issues/147

When using beta version of Postgres (e.g `17beta3`) the current code using `server_version` would fail. Using `server_version_num` overcome this and avoid having to parse the version number

### DIFF
--- a/src/pypgstac/python/pypgstac/migrate.py
+++ b/src/pypgstac/python/pypgstac/migrate.py
@@ -122,8 +122,16 @@ class Migrate:
             logger.info("using unreleased version")
             toversion = "unreleased"
 
-        pg_version = self.db.pg_version
-        logger.info(f"Migrating PgSTAC on PostgreSQL Version {pg_version}")
+        major, minor, patch = tuple(
+            map(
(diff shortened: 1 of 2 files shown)


src/pypgstac/python/pypgstac/migrate.py
```diff
@@ -122,8 +122,8 @@ def run_migration(self, toversion: Optional[str] = None) -> str:
             logger.info("using unreleased version")
             toversion = "unreleased"
 
-        pg_version = self.db.pg_version
-        logger.info(f"Migrating PgSTAC on PostgreSQL Version {pg_version}")
+        pg_version = int(self.db.pg_version)
+        logger.info(f"Migrating PgSTAC on PostgreSQL Version {pg_version/1000:,}")
```
vincentsarago, line 126: would print something like 
```
Migrating PgSTAC on PostgreSQL Version 17.0
```

bitner (approved):

PR #310, opened 2024-10-10
add tests for collections search links

vincentsarago:
while working on https://github.com/stac-utils/stac-fastapi-pgstac/pull/155 we encountered some strange behaviour with the next/prev links returned by pgstac

The first commit of this PR adds tests showing what I was expecting from pgstac 

cc @bitner 

Note: in tipg this is how we determine next/prev links https://github.com/developmentseed/tipg/blob/ea1ef949873abb7886811dce85a47913f38e3e38/tipg/collections.py#L783-L788


vincentsarago:
🤔 well tests pass!

vincentsarago:
@bitner I'm not sure to understand how the CI is design, thus I'm not sure that pgtap was ran with my last commit

bitner:
Yeah, it doesn't look like tests ran at all.

bitner (approved):

drnextgis:
In 'Files changed' tab I see changes that were already merged to the `main` branch, like this https://github.com/stac-utils/pgstac/commit/b73d5703e4b179cdaad185d6795b66efa3bc7087.

drnextgis:
Ah that makes sense because it was never included in the migrations.

PR #342, opened 2025-01-27
Typing improvements and loosened dependencies

### LINKED ISSUE #341: Loosen dependencies requirements
https://github.com/stac-utils/pgstac/blob/main/src/pypgstac/pyproject.toml

Such pinning makes it impossible to integrate this tool into other projects (without hacky workarounds like override-dependencies, forceful dependency overrides). And I assume the intention is to allow such use cases given the existence of [py.typed](https://github.com/stac-utils/pgstac/blob/main/src/pypgstac/src/pypgstac/py.typed) file which otherwise would be redundant.

There exist online resources on this topic: https://discuss.python.org/t/should-i-be-pinning-my-dependencies/13159

A popular and good practice is to use `>=` for defining your dependencies.

Zaczero:
This PR is a set of several typing improvements as well as loosened dependencies which resolves #341. While this PR bundles several changes, the commits are atomic and should be simple to review individually.

More notable changes:
- pydantic depdendency was replaced with pydantic-settings as in [official docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/).
- set_json_dumps no longer uses a .decode() wrapper method, since bytes return is supported in psycopg since 3.1.9.
- pathlib.Path type hints were removed since from my testing they never worked (they are not an instance of str).

### DIFF
--- a/src/pypgstac/pyproject.toml
+++ b/src/pypgstac/pyproject.toml
@@ -48,7 +48,7 @@ dev = [
     "pre-commit",
 ]
 psycopg = [
-    "psycopg[binary]>=3.1",
+    "psycopg[binary]>=3.1.9",
     "psycopg-pool>=3.1",
 ]
 migrations = [
--- a/src/pypgstac/pyproject.toml
+++ b/src/pypgstac/pyproject.toml
@@ -22,36 +22,34 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "cachetools==5.3.*",
-    "fire==0.4.*",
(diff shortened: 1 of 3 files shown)


bitner:
Hey @Zaczero Thanks for kicking this PR up here.

I am not too keen on just using >= for dependencies as (particularly with a major release version) there is a good chance that there could be breaking changes from dependencies.

For pydantic, I do know that we need to integrate with some libraries that still are using pydantic v1. This is why we use the v1 compatibility layer in Pydantic v2 that allows us to still use the old pydantic base settings api.

PR #352, opened 2025-02-21
deps: relax

### LINKED ISSUE #341: Loosen dependencies requirements

gadomski:
@bitner I'm of the `>=` school. If a upstream breaks, we can just ceil (CI should catch the break). I've added a test to CI to test with both our min version and whatever pip pulls in on a normal install.

- Closes #341

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -102,6 +102,11 @@ jobs:
     name: test
     needs: [changes, buildpg, buildpyrust]
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flags:
+          - ""
+          - "--resolution lowest-direct"
     container:
       image: ${{ needs.changes.outputs.pyrustdocker }}
       env:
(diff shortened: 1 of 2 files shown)


bitner (approved):

Zaczero:
`psycopg = ["psycopg[binary]==3.1.*", "psycopg-pool==3.1.*"]`

I think this selector should be loosed up as well. It's common for users to use it as well.

PR #361, opened 2025-03-23
Add `load_queryables` function to pypgstac

emmanuelmathot:
## Description

This PR adds a new `load_queryables` function to pypgstac that allows loading queryables from a JSON file. Queryables are a mechanism that allows clients to discover what terms are available for use when writing filter expressions in a STAC API.

The new function:
- Takes a JSON file path and optional collection IDs as input
- Parses the queryables definition from the JSON file
- Determines the appropriate property wrapper based on the type of each property
- Inserts the queryables into the database
- Handles conflicts with existing queryables by deleting and reinserting them to ensure triggers are executed

## Changes

- Added `load_queryables` method to the `PgstacCLI` class in `src/pypgstac/src/pypgstac/pypgstac.py`
- Updated documentation in `docs/src/pypgstac.md` to include information about the new command
- Created example files to demonstrate usage:
  - `src/pypgstac/examples/sample_queryables.json`: A sample queryables definition file
  - `src/pypgstac/examples/load_queryables_example.py`: An example script showing how to use the function programmatically
- Added comprehensive unit tests in `src/pypgstac/tests/test_queryables.py`
- Updated CHANGELOG.md to include the new feature

## Usage

The new command can be used as follows:

```bash
# Load queryables for all collections
pypgstac load_queryables queryables.json

# Load queryables for specific collections
pypgstac load_queryables queryables.json --collection_ids collection1,collection2
```

Or programmatically:

```python
from pypgstac.pypgstac import PgstacCLI

cli = PgstacCLI()
cli.load_queryables("queryables.json", collection_ids="collection1,collection2")
```

## Testing

The implementation has been tested with various scenarios:
- Loading queryables for all collections
- Loading queryables for specific collections
- Updating existing queryables
- Handling invalid JSON files
- Handling files with no properties

All tests pass successfully.

### DIFF
--- a/src/pypgstac/src/pypgstac/pypgstac.py
+++ b/src/pypgstac/src/pypgstac/pypgstac.py
@@ -119,6 +119,121 @@ class PgstacCLI:
                 except Exception:
                     pass
 
+    def load_queryables(
+        self,
+        file: str,
+        collection_ids: Optional[str] = None,
+    ) -> None:
+        """Load queryables from a JSON file.
+
+        Args:
+            file: Path to the JSON file containing queryables definition
+            collection_ids: Comma-separated list of collection IDs to apply the
+                            queryables
+                to
+        """
+        import orjson
+
+        from pypgstac.load import read_json
+
+        # Parse collection_ids if provided
+        coll_ids_array = None
+        if collection_ids:
+            coll_ids_array = [cid.strip() for cid in collection_ids.split(",")]
+
+        # Read the queryables JSON file
+        queryables_data = None
+        for item in read_json(file):
+            queryables_data = item
+            break  # We only need the first item
+
+        if not queryables_data:
+            raise ValueError(f"No valid JSON data found in {file}")
+
+        # Extract properties from the queryables definition
+        properties = queryables_data.get("properties", {})
+        if not properties:
+            raise ValueError("No properties found in queryables definition")
+
+        conn = self._db.connect()
+        with conn.cursor() as cur:
+            with conn.transaction():
+                # Insert each property as a queryable
+                for name, definition in properties.items():
+                    # Skip core fields that are already indexed
+                    if name in (
+                        "id",
+                        "geometry",
+                        "datetime",
+                        "end_datetime",
+                        "collection",
+                    ):
+                        continue
+
+                    # Determine property wrapper based on type
+                    property_wrapper = "to_text"  # default
+                    if definition.get("type") == "number":
+                        property_wrapper = "to_float"
+                    elif definition.get("type") == "integer":
+                        property_wrapper = "to_int"
+                    elif definition.get("format") == "date-time":
+                        property_wrapper = "to_tstz"
+                    elif definition.get("type") == "array":
+                        property_wrapper = "to_text_array"
+
+                    # Determine index type (default to BTREE)
+                    property_index_type = "BTREE"
+
+                    # First delete any existing queryable with the same name
+                    if coll_ids_array is None:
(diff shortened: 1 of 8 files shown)


src/pypgstac/src/pypgstac/pypgstac.py
```diff
@@ -119,6 +119,121 @@ def loadextensions(self) -> None:
...
+
+        Args:
+            file: Path to the JSON file containing queryables definition
+            collection_ids: Comma-separated list of collection IDs to apply the
+                            queryables
+                to
+        """
+        import orjson
+
+        from pypgstac.load import read_json
```
vincentsarago, line 137: can we move the `import` at the top level?

src/pypgstac/src/pypgstac/pypgstac.py
```diff
@@ -119,6 +119,121 @@ def loadextensions(self) -> None:
                 except Exception:
                     pass
 
+    def load_queryables(
+        self,
+        file: str,
+        collection_ids: Optional[str] = None,
```
vincentsarago, line 125: it seems that `fire` can understand the `list` type of so in theory you could pass `pypgstac load_queryables ... --collections_ids=["collection1", "collection2"]`

ref: https://github.com/google/python-fire/blob/master/docs/guide.md#argument-parsing

src/pypgstac/src/pypgstac/pypgstac.py
```diff
@@ -119,6 +119,121 @@ def loadextensions(self) -> None:
...
+                    elif definition.get("format") == "date-time":
+                        property_wrapper = "to_tstz"
+                    elif definition.get("type") == "array":
+                        property_wrapper = "to_text_array"
+
+                    # Determine index type (default to BTREE)
+                    property_index_type = "BTREE"
+
+                    # First delete any existing queryable with the same name
+                    if coll_ids_array is None:
```
vincentsarago, line 188: I think you'll want `if not coll_ids_array` here to also catch empty list

src/pypgstac/src/pypgstac/pypgstac.py
```diff
@@ -119,6 +120,175 @@ def loadextensions(self) -> None:
...
+                        property_wrapper = "to_float"
+                    elif definition.get("type") == "integer":
+                        property_wrapper = "to_int"
+                    elif definition.get("format") == "date-time":
+                        property_wrapper = "to_tstz"
+                    elif definition.get("type") == "array":
+                        property_wrapper = "to_text_array"
+
+                    # Determine index type (default to BTREE)
+                    property_index_type = "BTREE"
```
bitner, line 181: I'm leery about always setting an index. Adding too many indexes can actually be far worse than not having an index in many situations. It will likely lead to worse performance to have an index on every possible field. It will definitely make any write operations much slower and will lead to memory contention in many instances on read operations. Perhaps if we add another option indexfields=[foo, bar] in order to define which fields to add indexes to???

bitner (changes requested):
As per my comment above, I think that we should not create an index for every possible queryable. That is likely to degrade performance more than it would help. Indexes should only be created on the most used or most selective queryables. I think adding a index_fields argument that allows for giving a list of fields which indexes should be created for (Or possibly a second command just for setting indexing???)

src/pypgstac/src/pypgstac/pypgstac.py
emmanuelmathot: Fine. What should we put in `property_wrapper` and `property_index_type` for not having the index created?

bitner (approved):

Issue #328, opened 2024-12-03
pyPgSTAC still uses `pipes` module despite it being deprecated in Python 3.11


GeoSpark:
`pipes` was deprecated in Python 3.11 and removed in 3.13 See: https://docs.python.org/3.13/library/pipes.html

bitner:
@GeoSpark I don't believe that we are directly using that library, if it is getting installed, it must be in one of our dependencies. I'm not going to hunt that down, but I'm happy to review any PRs!

luipir:
FYI 

18:20 $ pypgstac loads
Traceback (most recent call last):
  File "./envs/default/bin/pypgstac", line 5, in <module>
    from pypgstac.pypgstac import cli
  File "./envs/default/lib/python3.13/site-packages/pypgstac/pypgstac.py", line 7, in <module>
    import fire
  File "./envs/default/lib/python3.13/site-packages/fire/__init__.py", line 21, in <module>
    from fire.core import Fire
  File "./envs/default/lib/python3.13/site-packages/fire/core.py", line 59, in <module>
    import pipes
ModuleNotFoundError: No module named 'pipes'

PR #362, opened 2025-03-27
Udpate fire to 0.7.0

drnextgis:
Fixes https://github.com/stac-utils/pgstac/issues/328. The [fix](https://github.com/google/python-fire/pull/447) was included in the 0.7.0 release.

### DIFF
--- a/src/pypgstac/pyproject.toml
+++ b/src/pypgstac/pyproject.toml
@@ -19,7 +19,7 @@ classifiers = [
 ]
 dependencies = [
     "cachetools>=5.3.0",
-    "fire>=0.4.0",
+    "fire>=0.7.0",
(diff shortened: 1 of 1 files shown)


vincentsarago:
🙏 thanks @drnextgis I wanted to take care of this this week as well 🙏 

the update to fire 0.7.0 will enable python3.13 🥳 

I'm thinking we could maybe refactor a bit https://github.com/stac-utils/pgstac/blob/905e05b2d72aaf0d6e04236713371f0b42a1c468/.github/workflows/continuous-integration.yml#L132-L134 to tests multiple version of python but that can be done in another PR 🙏

bitner (approved):

PR #370, opened 2025-05-12
Add `--create-missing-collections` option to load_queryables

emmanuelmathot:
This PR adds a new option `--create-missing-collections` to the `load_queryables` command that automatically creates missing collections rather than failing.

Changes:
- Added new optional parameter `create_missing_collections` 
- When enabled, non-existent collections are created with valid STAC metadata
- Uses proper transaction handling for database operations
- Added comprehensive test coverage
- Created collections include required STAC fields and global extent

### DIFF
--- a/src/pypgstac/src/pypgstac/pypgstac.py
+++ b/src/pypgstac/src/pypgstac/pypgstac.py
@@ -126,6 +126,7 @@ class PgstacCLI:
         collection_ids: Optional[list[str]] = None,
         delete_missing: Optional[bool] = False,
         index_fields: Optional[list[str]] = None,
+        create_missing_collections: Optional[bool] = False,
     ) -> None:
         """Load queryables from a JSON file.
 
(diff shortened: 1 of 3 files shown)


src/pypgstac/src/pypgstac/pypgstac.py
```diff
@@ -147,6 +151,59 @@ def load_queryables(
             queryables_data = item
             break  # We only need the first item
 
+            # Create missing collections if requested
```
vincentsarago, line 154:
```suggestion
        # Create missing collections if requested
```

bitner: @emmanuelmathot I think it makes sense to have a generic SQL create_empty_collections function. Right now, you've got some slight SQL injection vulnerabilities in there (f strings and SQL are not good friends). Do you mind if I push some changes to this PR?

PR #375, opened 2025-07-09
refactor pgstacDb class

### LINKED ISSUE #340: Database connections are never closed
PgstacDB maintains a persistent connection to the database through `self.connection = pool.getconn()`. According to the `getconn()` documentation, this connection will not be automatically returned to the pool, and therefore won't be closed when idle. An explicit `putconn()` call is required. Currently, `db_max_idle` is ignored, which leads to database timeouts during long idle periods. I suggest not implementing a custom `self.connection` singleton pooling logic, but rather relying on the `psycopg_pool` implementation. Otherwise, you would need to implement the same feature set as `psycopg_pool`, including idle handling, which adds unnecessary complexity.

vincentsarago:
This PR tries to resolve #340 by completely refactoring the PgstacDb class and its usage by other classes/methods

### DIFF
--- a/src/pypgstac/src/pypgstac/db.py
+++ b/src/pypgstac/src/pypgstac/db.py
@@ -107,6 +107,8 @@ class PgstacDB:
     def connect(self) -> Connection:
         """Return database connection."""
         pool = self.get_pool()
+
+        conn: Connection = None
         try:
             conn = pool.getconn()
             conn.autocommit = True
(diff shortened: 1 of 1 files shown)


src/pypgstac/src/pypgstac/db.py
```diff
@@ -52,37 +54,24 @@ class Settings(BaseSettings):
...
-        self.commit_on_exit = commit_on_exit
-        self.initial_version = "0.1.9"
-        self.debug = debug
-        self.use_queue = use_queue
-        if self.debug:
-            logging.basicConfig(level=logging.DEBUG)
+    dsn: str
+    commit_on_exit: bool = True
+    debug: bool = False
+    use_queue: bool = False
```
vincentsarago, line 64: removed possibility to pass a `Connection` object which adds to much complexity

PR #379, opened 2025-08-04
allow array for q full text search

bitner:
Allow the paramter "q" for full text search to be an array as well as text.


vincentsarago (approved):
😍

PR #387, opened 2025-09-29
fix: combine space-separated terms with adjacency operator in free-text search

hrodmn:
There is some ambiguity between the free-text STAC API extension and the OGC Records specification (https://github.com/stac-api-extensions/freetext-search/issues/12), so I am not 100% sure this is the right thing to do, but right now if you pass a string with two terms separated by spaces (but not quoted), pgstac will throw a tsquery syntax error (https://github.com/stac-utils/pgstac/actions/runs/18101825325/job/51506541624).

From the [OGC Records Spec](https://docs.ogc.org/is/20-004r1/20-004r1.html#core-query-parameters-q):
> For multiple search terms that are white space separated, only records that contain all the search terms specified, in the order specified and separated by any number of white spaces in one or more of the searched text fields SHALL be in the result set.

This PR replaces spaces with ~~`&`~~ `<->` which I think is the behavior that most closely aligns with the specification. ~~but does not account for the order of the terms like the spec suggests. I tried using distance operators (e.g. `<10>`) but that checks for an exact distance between terms ("distance equals 10") instead of checking for "less than or equal to 10".~~

### DIFF
--- a/src/pgstac/migrations/pgstac.0.9.8-unreleased.sql
+++ b/src/pgstac/migrations/pgstac.0.9.8-unreleased.sql
@@ -280,10 +280,10 @@ BEGIN
     processed_text := regexp_replace(processed_text, '^\s*\-([a-zA-Z0-9_]+)', '! \1', 'g'); -- -term at start
     processed_text := regexp_replace(processed_text, '\s*\-([a-zA-Z0-9_]+)', ' & ! \1', 'g'); -- -term elsewhere
 
-    -- terms separated with spaces are assumed to represent an AND clause. loop through these
-    -- occurrences and replace them with &
+    -- terms separated with spaces are assumed to represent adjacent terms. loop through these
+    -- occurrences and replace them with the adjacency operator (<->)
     LOOP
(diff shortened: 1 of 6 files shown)


vincentsarago:
re-reading the spec, `word1 word2` doesn't mean `word1 AND word2`. Could we just `quote` the expression?

hrodmn:
> re-reading the spec, `word1 word2` doesn't mean `word1 AND word2`. 

It depends if you are looking at the OGC Records spec or the free-text search STAC API extension spec :sweat_smile:. I *think* the collection-search extension means to be based on the OGC Records spec but it is not correctly interpreting that particular case (see https://github.com/stac-api-extensions/freetext-search/issues/12 for discussion). 

I think the OGC spec does mean space-separated words like `word1 word2` to be `word1 AND word2`: "For multiple search terms that are white space separated, **only records that contain all the search terms specified**, in the order specified and separated by any number of white spaces in one or more of the searched text fields SHALL be in the result set."

> Could we just quote the expression?

Yeah, that would probably be pretty close to the same thing as using the distance operator I think, but might give users false negatives if there are indeed extra whitespaces between the terms in the text fields.

Issue #158, opened 2023-02-05
search response does not include datetime when null


duckontheweb:
## Description

The STAC Spec [requires that Items have a `datetime` property](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#properties-object) even when the value of that property is `null`. However, in the return value of the `search` function removes the `datetime` property when it is null, resulting in an invalid STAC Item.

## Steps to reproduce

1. Start up the development server
2. In a `psql` console, add an item with `start_datetime` and `end_datetime` values and a `null` `datetime` value:
   ```sql
   SELECT *
   FROM create_item('
   {
     "type": "Feature",
     "stac_version": "1.0.0",
     "id": "datetime_range_test",
     "properties": {
        "start_datetime": "2009-10-15T00:00:00Z",
        "end_datetime": "2010-03-01T23:59:59.999999Z",
        "datetime": null
      },
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [-101, 36],
            [-101, 37],
            [-102, 37],
            [-102, 36],
            [-101, 36]
          ]
        ]
      },
      "links": [],
      "assets": {},
      "collection": "pgstac-test-collection"
   }
   '::jsonb);
   ```
3. Use the `search` function to fetch the Item and examine the properties:
   ```sql
   SELECT "search" -> 'features' -> 0 -> 'properties'
   FROM search('{"ids": ["datetime_range_test"]}');
   ```

The resulting `properties` JSON is missing the `datetime` property:

```json
{
  "end_datetime": "2010-03-01T23:59:59.999999Z",
  "start_datetime": "2009-10-15T00:00:00Z"
}
```

bitner:
Per the spec in the section you linked to "null is allowed, but requires start_datetime and end_datetime from [common metadata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time-range) to be set."

My reading of that and the implementation in pgstac is that an item must EITHER have datetime OR BOTH start_datetime and end_datetime. If there has been a change in the spec or I have been reading that wrong, please reopen this issue, but I am going to close it now.

duckontheweb:
> Per the spec in the section you linked to "null is allowed, but requires start_datetime and end_datetime from [common metadata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time-range) to be set."
> 
[...]

PR #391, opened 2025-11-19
Add datetime: null when an item has start_datetime and end_datetime but no datetime.

krisaoe:
Ensures STAC spec compliance by adding datetime: null when temporal ranges are present. See #158


gadomski (changes requested):
It'd be good to add a test to confirm the expected behavior. I'm also summoning @bitner in case he's got insight into the fix strategy.

krisaoe:
Moved the fix to the jsonb overload - all hydration paths flow through it (including tests).
Added 2 pgtap tests (adds datetime key when start/end_datetime present & preserves existing datetime), added a hydration basic test to test the full flow.

bitner:
@krisaoe the only things that we strip out of an items json on ingest are the id, geometry, collection, and type. See the content_slim function in 003a_items.sql. We leave (correctly or incorrectly) datetime, start_datetime, and end_datetime exactly as they were in the original json. In very early versions, we stripped out those fields and reconstructed them from the instantiated datetime and end_datetime fields in the database on output, but there were instances where when we re-put those back into the output json that there were issues with null handling (as you are having) or with precision, so we made the decision then that we would keep those fields within the json (at the expense of duplicating that storage) to make sure that things were faithfully round tripped.

PGStac does not try to do any validating of stac records and expects that any validation is done as part of an ingest workflow.

Can you verify if the behavior I am describing above is what you are seeing or if you are seeing something come out of pgstac with json that does not match what had been ingested into pgstac?

vincentsarago:
FYI there was a bug in stac-pydantic/stac-fastapi which was stripping null datetime, but this is now resolved https://github.com/stac-utils/stac-pydantic/pull/190

PR #421, opened 2026-04-10
pypgstac: lift version constraint on cachetools

vprivat-ads:
The current version restriction forbids to use pypgstac alongside recent versions of cachetools ([7.0.5](https://pypi.org/project/cachetools/7.0.5/) right now)


vprivat-ads:
@bitner this is a partial revert of the restriction added in https://github.com/stac-utils/pgstac/pull/408, I don't understand why it has been added there?

vprivat-ads:
I see the change is being reverted in #427 so I guess this PR can be closed?

Issue #425, opened 2026-04-13
bug: `jsonb_strip_nulls()` in hydration functions strips datetime: null from item properties, producing invalid STAC items


seedlit:
pgstac's database level hydration drops `"datetime": null` from STAC item properties when reconstructing items for API responses. This produces invalid STAC items when `start_datetime`/`end_datetime` are used, because the STAC spec requires `"datetime": null` to be [explicitly present in that case.](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#datetime)

### 1. Ingest an item with `datetime: null`

Ingest any STAC item that uses a date range (i.e., `datetime` is `null` with `start_datetime` and `end_datetime` set).

### 2. Query the item back via the search function

```sql
SELECT content_hydrate(items) FROM items WHERE id = '<your-item-id>';
```

Or via stac-fastapi-pgstac:

```bash
curl -s 'http://localhost:8081/collections/<collection>/items?limit=1' | python3 -c "
import json, sys
data = json.load(sys.stdin)
print(json.dumps(data['features'][0]['properties'], indent=2))
"
```

### 3. Observe that `datetime` is missing

**Expected:**
```json
{
  "datetime": null,
  "start_datetime": "2024-01-01T00:00:00Z",
  "end_datetime": "2024-01-02T00:00:00Z"
}
```

**Actual:**
```json
{
  "start_datetime": "2024-01-01T00:00:00Z",
  "end_datetime": "2024-01-02T00:00:00Z"
}
```

`"datetime"` is entirely absent from the response, not just null.

seedlit:
this issue is duplicate of https://github.com/stac-utils/pgstac/issues/158

PR #436, opened 2026-05-09
fix: preserve "datetime": null in hydration/dehydration (jsonb_strip_nulls strips nested nulls)

seedlit:
## Summary

Addresses https://github.com/stac-utils/pgstac/issues/425

Fixes a bug where pgstac's database-level hydration drops `"datetime": null` from STAC item properties, producing STAC-invalid items when `start_datetime`/`end_datetime` are used.

Per the [STAC Item spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#datetime), when a single datetime is not appropriate, the `datetime` field **must** be set to `null` (not absent) and `start_datetime`/`end_datetime` must be supplied.


seedlit:
addressed by https://github.com/stac-utils/pgstac/pull/448

PR #440, opened 2026-05-12
Harden search cache and add pg_stat_statements/pg_cron support

bitner:
## Summary
This branch tightens PgSTAC search cache behavior, improves concurrency safety around search stats, and adds first-class support for `pg_stat_statements` and `pg_cron` in the pgstac Docker image and test flow.

## What changed
- Reworked search cache handling to use a canonical where-clause hash and better concurrency controls when creating and touching cached searches.
- Added search stats refresh support so `updatestats` flows through cached and uncached search paths and keeps `numberMatched` / context counts current.
- Add columns and tooling to be able to save named queries, to pin queries so they won't be cleaned by TTL values, and to have garbage collection that can clean up not-recently used, named, or pinned queries.
- Add garbage collection function for searches (this still must be called by something such as pg_cron).
- Added `pg_stat_statements` and `pg_cron` to the pgstac image, enabled them through `shared_preload_libraries`, and initialized them during container bootstrap.
- Added smoke tests for both extensions in `scripts/container-scripts/test`.
- Updated pgtap/basic SQL coverage around search, token handling, readonly behavior, and related cache behavior.

## Why
The goal of the search-cache work is to reduce collision risk, avoid stale or inconsistent stats under concurrent requests, and make the search implementation easier to maintain by keeping hashing and cache logic in the same SQL module. We also want to be able to clear out the cache for any searches that we don't need to actively be able to look up by name or hash (for example for titler-pgstac integration).

The extension work makes the standard test image closer to the production runtime and gives us direct verification that both extensions are loaded and usable, rather than assuming the container configuration is correct.


gadomski (approved):
Do we need to update the best-practices docs to tell folks to set up a cronjob to clean up searches?

bitner:
> Do we need to update the best-practices docs to tell folks to set up a cronjob to clean up searches?

At the end of this series of PRs, there's going to need to be a big clean up of docs including several "cron ready" functions/procedures that I'd rather document all together.

PR #441, opened 2026-05-13
Optimize ingestion and metadata search lifecycle

bitner:
Adding tracking to tables to enable synchronization tracking between pgstac and external sources. This includes adding a hash of each item to be able to diff between local sources and pgstac to limit the bandwidth when running updates.

Add deletion table with garbage collection to be able to track what records have been removed for use in synchronization.


gadomski (approved):
Makes sense to me!

PR #444, opened 2026-05-20
pgstac sql changes for updated table layout

bitner:


scripts/container-scripts/test
```diff
@@ -468,8 +469,9 @@ then
     SETUPDB=1
     PGTAP=1
     BASICSQL=1
-    PYPGSTAC=1
-    MIGRATIONS=1
+    # PYPGSTAC and MIGRATIONS are intentionally excluded from the default
+    # run-all block while Python loader updates are pending on this branch.
+    # Use --pypgstac or --migrations flags to run them explicitly.
     PGDUMP=1
 fi
```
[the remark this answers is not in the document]
bitner: this is ok, we are just disabling these for now as we will work on pypgstac and the incremental migrations later in this process to create the v0.10.0 release

src/pgstac/sql/001a_jsonutils.sql
```diff
@@ -147,25 +147,31 @@ CREATE OR REPLACE FUNCTION jsonb_fields(j jsonb, f jsonb DEFAULT '{"fields":[]}'
...
 
 
+-- merge_jsonb: Deep-merge two JSONB values, with _a taking precedence over _b.
+-- Null semantics (v0.10+):
+--   SQL NULL _a  → return _b (no information from _a, use _b as default)
+--   JSON null _a → return JSON null (explicit null wins; STAC datetime:null must survive)
+--   '"𒍟※"'   _a  → return SQL NULL (sentinel used by strip_jsonb to mark removed values)
+-- Objects are merged key-by-key recursively; same-length arrays are merged element-by-element.
+-- Any other type: _a wins.
 CREATE OR REPLACE FUNCTION merge_jsonb(_a jsonb, _b jsonb) RETURNS jsonb AS $$
```
bitner: @copilot do we still need merge_jsonb and strip_jsonb with the new fragments setup? or at least, do we need to modify things so that we aren't using the null marker any more? are there any further optimizations that we can do here?

src/pgstac/sql/002_collections.sql
```diff
@@ -1,36 +1,44 @@
...
-        'type', 'Feature',
-        'stac_version', content->'stac_version',
-        'assets', content->'item_assets',
-        'collection', content->'id'
-    );
-$$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
+-- collection_fragment_config_default: Derive a fragment_config text[] from item_assets if present.
+-- Returns one serialized path per item_asset key (e.g. 'assets.thumbnail'), or NULL if no item_assets.
+-- This is used when creating/upserting a collection to auto-populate fragment_config.
+CREATE OR REPLACE FUNCTION collection_fragment_config_default(content jsonb) RETURNS text[] AS $$
```
bitner, line 11: @copilot we need to make sure that we have a good default here. this should include ignoring base fields like stac_extensions and stac_version and any other fields like the proj fields that are likely to be the same for every item in a collection and for things in item_assets, it should look for things that are not going to change from item to item, but we should make sure not to include things like href that will be different per item. look at various collections and items on the microsoft planetary computer (https://planetarycomputer.microsoft.com/api/stac/v1/openapi.json) and come up with a plan so that we can more smartly identify a good base set of paths to allow stripping that highly duplicated data.

src/pgstac/sql/002_collections.sql
```diff
@@ -1,36 +1,44 @@
...
     partition_trunc text CHECK (partition_trunc IN ('year', 'month'))
 );
 
-CREATE OR REPLACE FUNCTION collection_base_item(cid text) RETURNS jsonb AS $$
-    SELECT pgstac.collection_base_item(content) FROM pgstac.collections WHERE id = cid LIMIT 1;
-$$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
-
-
+-- create_collection: Insert a new collection and derive fragment_config from item_assets.
 CREATE OR REPLACE FUNCTION create_collection(data jsonb) RETURNS VOID AS $$
```
bitner: @copilot, let's provide an overload that defaults to null to specify the partition_trunc and another one to specify the fragment_config for this function and the other crud functions here

src/pgstac/sql/002a_queryables.sql
bitner, line 126: @copilot does it still make sense to have this function that returns all the separate outs or should we break this into smaller functions that each only return a single output. it seems like some of the other functions introduced here might already be working towards that

src/pgstac/sql/998_idempotent_post.sql
```diff
@@ -25,6 +25,112 @@ DO $$
   END
 $$;
 
+-- Register promoted native-column queryables (v0.10 split schema).
+-- Each entry maps a STAC property name to the promoted items column via property_path.
+-- CQL2 queries and auto-created indexes will use the native column, not JSONB extraction.
+-- First block: insert new rows only. Second block below: update existing rows that have
+-- property_path=NULL (handles databases upgraded from pre-v0.10 without re-installing).
+WITH promoted_queryables(name, definition, property_path, property_wrapper) AS (
```
bitner: @copilot we are repeating ourselves a lot here, can we do something so that we only specify this values list once?

PR #448, opened 2026-06-01
Item fragments

### LINKED ISSUE #158: search response does not include datetime when null

### LINKED ISSUE #425: bug: `jsonb_strip_nulls()` in hydration functions strips datetime: null from item properties, producing invalid STAC items

bitner:
## Technical Context
- **Related Issues:** Fixes #158, Fixes #425
- **Breaking Changes:** Yes — complete replacement of the `items` table schema. The monolithic `content jsonb` column is gone. Common STAC properties and properties from extensions currently marked as stable are promoted into actual columns. Hydration/Dehydration still exists, but is now tied to a versioned item_fragments table rather than directly to item_assets on the collections table. 

## Description

Reworks how STAC items are stored and retrieved for the upcoming v0.10.0 breaking release of pgstac. 
**New storage model**

- `item_fragments` — deduplicated shared subtrees (asset metadata, link shapes, root keys) stored once per collection, keyed by a 32-byte sha256 (`hash bytea`). Items reference their fragment via `fragment_id`.
- `items` — per-item delta columns (`assets`, `properties`, `links`, `extra`) plus ~30 promoted scalar columns for well-known queryables (`datetime`, `platform`, `gsd`, `eo:*`, `proj:*`, `view:*`, `sat:*`, `file:*`, `sci:*`). 
- `collections.fragment_config text[]` — list of fragment paths, auto-derived from `item_assets` on collection creation. This can be overridden to be able to further optimize and deduplicate information stored across items that is common across all items in a collection.
- `items.link_hrefs` / `item_fragments.links_template` — link split storage: shared link shape (rel/type/title) deduped into the fragment; per-item hrefs stored separately.
- `item_field_registry` — tracks observed JSON paths per collection for queryable discovery. This will also allow figuring out the full schema of all data in a collection for use when storing to schema-requiring formats like parquet.

Items_staging tables have been updated to work with schema changes.

Functions are added to create a canonical hash that can be calculated the same internally in postgres/pgstac as well as externally to allow for fast lookups/diffing when loading data.

**Fixes #158 and #425 — `datetime: null` round-trip**

The STAC spec requires `"datetime": null` to be explicitly present when `start_datetime`/`end_datetime` are used. Earlier versions applied `jsonb_strip_nulls` to the full properties object, silently dropping it. The new `temporal_properties_from_item` builds `jsonb_build_object('datetime', NULL)` before the `jsonb_strip_nulls` block that covers only promoted scalars, so the explicit JSON null survives end-to-end through both `get_item` and `search`.

**Test gate**

`scripts/test --formatting --pgtap --basicsql --pgdump` is green (349 PGTap tests; pg_dump → `pgstac_restore` round-trip verified).

`--pypgstac` and `--migrations` are **intentionally skipped** These will be fixed in upcoming PRs prior to pgstac v0.10.0 release.  We are intentionally keeping the slices for PRs leading to v0.10.0 smaller and allowing for some tests not to pass to allow us to iterate to the point that we can be ready for this breaking release.

---

## Checklist
- [x] **Linting:** Code is formatted and linted
- [x] **Tests:** Tests pass. New tests added for split-storage round-trip, fragment dedup, canonical hash contract, datetime:null fix (#158/#425), jsonb_merge_recursive depth-4/collision correctness, and item_hash bytea sizing.
- [x] **Edge Cases:** Verified: depth-4 fragment paths, scalar key collisions in merge, empty items, range items with datetime:null via both get_item and search, upsert idempotency (re-upsert 1,000 items with no fragment explosion), changed-upsert reflected, direct UPDATE leaves item_hash stable.
- [ ] **Documentation:** README not updated (no new env vars or config); CLAUDE.md, AGENTS.md, and CHANGELOG.md updated.
- [x] **Accountability:** I can explain the implementation logic for every line of code submitted.

---

## AI tool usage
- [x] AI (Copilot or something similar) supported my development of this PR. See our [policy about AI tool use](https://stac-utils.github.io/ai-contribution-policy). Use of AI tools *must* be indicated.

---

> **Policy:** We require a "human-in-the-loop." You are the author and are fully accountable for all submitted code. Please ensure all tool-generated content is thoroughly reviewed before submission to ensure it is not an "extractive contribution" that squanders maintainer time.


bitner:
@copilot turn off the rustac tests in CI, we do not expect these to work yet with the schema rework.

src/pgstac/sql/001a_jsonutils.sql
```diff
@@ -79,6 +79,229 @@ CREATE OR REPLACE FUNCTION explode_dotpaths_recurse(IN j jsonb) RETURNS SETOF te
 $$ LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
 
 
+-- jsonb_canonical: RFC 8785 (JSON Canonicalization Scheme)-aligned serialization.
```
hrodmn, line 82: This is great!

src/pgstac/sql/002a_queryables.sql
```diff
@@ -567,3 +620,83 @@ CREATE OR REPLACE FUNCTION missing_queryables(_tablesample float DEFAULT 5) RETU
     ORDER BY 2,1
     ;
 $$ LANGUAGE SQL;
+
+-- promoted_item_property_defs: Shared STAC-property mapping for promoted native
+-- item columns. This function lives with queryables metadata so queryable seeding
+-- and items-table property extraction reference the same source of truth.
+CREATE OR REPLACE FUNCTION promoted_item_property_defs()
```
hrodmn, line 627: It would be nice to document somewhere the criteria that we would use when deciding promote additional fields.

src/pgstac/sql/002_collections.sql
```diff
@@ -1,50 +1,137 @@
...
+    END IF;
+
+    RETURN paths;
+END;
+$$ LANGUAGE PLPGSQL STABLE;
 
 CREATE TABLE IF NOT EXISTS collections (
     key bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
     id text GENERATED ALWAYS AS (content->>'id') STORED UNIQUE NOT NULL,
     content JSONB NOT NULL,
```
hrodmn, line 56: Maybe it is outside the scope of this PR but are you considering moving collection properties into top-level columns instead of stuffing them into `content`?

src/pgstac/sql/003a_items.sql
```diff
@@ -413,3 +908,274 @@ UPDATE collections
...
+-- The NOT EXISTS sub-select is evaluated per fragment; with an index on items.fragment_id
+-- this is an efficient anti-join rather than a full seq-scan.
+--
+-- Operational note: because items.fragment_id is intentionally unmanaged by an FK
+-- (partitioned-items incremental NOT VALID FKs are not supported), gc_fragments has a
+-- small race window with concurrent inserts. A fragment that is unreferenced at the time
+-- the DELETE snapshot is taken but becomes referenced by a later insert could be removed.
+-- The retention_interval guard makes this unlikely for normal ingest, but operators should
+-- still run gc_fragments during low-ingest periods or with a sufficiently conservative
+-- retention interval. This is a documented operational tradeoff, not a silent invariant.
```
hrodmn, line 1134: Thanks for the explanation about the lack of foreign key for `fragment_id`

src/pgstac/tests/basic/crud_functions.sql
```diff
@@ -18,44 +18,44 @@ INSERT INTO collections (content, partition_trunc) VALUES ('{"id":"pgstactest-cr
 
 -- Create an item
 SELECT create_item((SELECT content FROM test_items LIMIT 1));
-SELECT id, geometry, collection, datetime, end_datetime, content, private FROM items WHERE collection='pgstactest-crudtest';
```
hrodmn, line 21: Is `private` gone now?

hrodmn (commented):
The changes look good to me! Switching to the explicit top-level columns for item properties seems like a good directional change - are we going to do something similar for collections or does the small scale of the collections table not warrant the added complexity?

Note: I am trying to update my dev environment with `scripts/update` and `scripts/migrate` so I can interact with the new db schema but I can't seem to get the right incantation.

I tried:
```bash
scripts/update  # exits cleanly
scripts/migrate  # fails
```
error:
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "pypgstac": executable file not found in $PATH

```

src/pgstac/sql/002a_queryables.sql
bitner: "promoted" fields are still hard coded, there is no user ability to modify those. I chose the list based on common properties and the properties of extensions marked as "stable" - I would consider the "criteria" to be that on any major version release (since it is a breaking change to the schema) that we would update that list with anything new that has been marked "stable".

src/pgstac/sql/002_collections.sql
bitner: I wasn't because collections is generally small enough, that it's not really a "hot path", so the gains relative to items is pretty minimal.

src/pgstac/tests/basic/crud_functions.sql
bitner: no it's still there but it still is basically a place to stash things that all the mechanisms of pgstac ignore

bitner:
> The changes look good to me! Switching to the explicit top-level columns for item properties seems like a good directional change - are we going to do something similar for collections or does the small scale of the collections table not warrant the added complexity?
> 
> Note: I am trying to update my dev environment with `scripts/update` and `scripts/migrate` so I can interact with the new db schema but I can't seem to get the right incantation.
> 
> I tried:
> 
> ```shell
> scripts/update  # exits cleanly
> scripts/migrate  # fails
> ```
> 
> error:
> 
> ```
> Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "pypgstac": executable file not found in $PATH
> ```

So, the next couple releases with these SQL changes have broken some things that we will be working on later so that we could keep the amount to review per PR more reasonable, so right now there is no incremental migration and some things with pypgstac are broken. The incremental migration is going to be it's own PR at the end of this cycle of v0.10 development. Also, there will be a PR that replaces pypgstac (or more accurately, just makes it a wrapper around rustac pgstac), but the API will be a breaking change and will more closely reflect the rustac/pystac api. I'll update the PR description with instructions on how to get things set up in the meantime.

.github/workflows/continuous-integration.yml
```diff
@@ -184,6 +184,7 @@ jobs:
         run: scripts/container-scripts/test
   rust-crate:
     name: Test rust crate
+    if: ${{ false }}
```
gadomski: I like to leave FIXMEs so that I (or a robot) knows to come back later.

```suggestion
    if: ${{ false }}  # FIXME: turn back on before v0.10 release
```

docs/src/pgstac.md
```diff
@@ -85,12 +85,10 @@ Note that when pgstac.readonly is set to TRUE that pgstac is unable to use a cac
...
 
-Runtime configuration is available for **context**, **context_estimated_count**, **context_estimated_cost**, **context_stats_ttl**, and **nohydrate**.
+Runtime configuration is available for **context**, **context_estimated_count**, **context_estimated_cost**, and **context_stats_ttl**.
 
-The nohydrate conf item returns an unhydrated item bypassing the CPU intensive step of rehydrating data with data from the collection metadata. When using the nohydrate conf, the only fields that are respected in the fields extension are geometry and bbox.
-```sql
-SELECT search('{"conf":{"nohydrate"=true}}');
-```
+The legacy `conf.nohydrate` flag is still accepted in the request JSON for backward
+compatibility, but split-storage search always returns hydrated items.
```
gadomski, line 91: Am I reading this correctly, that hydration is _always_ at the database layer now? If so, do we need to exercise this through **stac-fastapi-pgstac** to make sure this won't be a performance regression?

src/pgstac/sql/002_collections.sql
```diff
@@ -1,50 +1,137 @@
...
+    -- Fields within each asset that differ between items and must NOT be fragmented.
+    -- href is unique per item; file:* fields reflect per-file measurements;
+    -- alternate/storage:* are access-layer derived paths also unique per item.
+    per_item_asset_fields CONSTANT text[] := ARRAY[
+        'href',
+        'file:size', 'file:checksum', 'file:local_path',
+        'alternate',
+        'storage:path', 'storage:platform', 'storage:region',
+        'storage:requester_pays', 'storage:tier'
+    ];
```
gadomski, line 25: 🤔 if things vary, they shouldn't be `item_assets` anyways, right? E.g. `href` isn't in any of the values of `Collection.item_assets`.

src/pgstac/sql/002a_queryables.sql
gadomski: I feel a little uneasy about promoting extension fields — though the extension may be "stable", it might have breaking releases (e.g. proj). Core fields I'm much more comfortable with.

src/pgstac/sql/003a_items.sql
```diff
@@ -1,12 +1,96 @@
...
+    updated timestamptz,
+    platform text,
+    instruments text[],
+    constellation text,
+    mission text,
+    eo_cloud_cover float8,
+    eo_bands jsonb,
+    eo_snow_cover float8,
+    gsd float8,
+    proj_epsg integer,
```
gadomski: Yeah, this is exactly what I was worried about w.r.t. extensions. Current version of [projection](https://github.com/stac-extensions/projection#fields) doesn't have an `epsg` field, it changed to `code`.

gadomski (changes requested):
Pausing review mid-stream while we figure out extension fields in the item fragments.

docs/src/pgstac.md
bitner: no, but rather than being a flag to return an item collection that then needs to be blown apart and reconstructed, the new rustac pgstac (and python wrapper) functions will make a query that gets the raw rows (using functions coming in the next PR) and so those tools (which stac-fastapi-pgstac will need to be updated to use) will have a much faster and memory efficient path. we still maintain the search function though that returns the full item collection which will not have that option any longer though and so for things that use THAT function, yes, there might be a performance regression, but with that being said, the new hydration flow IS considerably faster than the old one.

src/pgstac/sql/002_collections.sql
bitner: Fair point, this is a bit of an artifact from iterating through a number of ideas for generating the default fragment config, but I think it's ok to leave in as a safety valve since there are no constraints on what anybody might stick into item_assets.

src/pgstac/sql/003a_items.sql
bitner: Hrmmm, yeah, this one might be worth hopping on a call and chatting through. Basically, any property that we can promote is going to have performance benefits for querying and for using the fields extension. It also has potential to reduce the storage overhead (by not storing the key name, but note I say potential as the internal TOAST compression in postgres could potentially counteract that benefit, but anything that is toasted does add that overhead when we access or filter based on that field, so storage space is not strictly the main goal of promoting fields).

I'm trying to balance here promoting as much as we can that we should expect to be reasonably stable.

gadomski: > I'm trying to balance here promoting as much as we can that we should expect to be reasonably stable.

My instinct is to default to core fields, but allow an installing user to add more?

Let's chat tomorrow.

src/pgstac/sql/002_collections.sql
gadomski: > there are no constraints on what anybody might stick into item_assets

That suggests a default allow list might be safer, and allow folks to improve performance by manually providing more fields?

src/pgstac/sql/003a_items.sql
hrodmn: would it be reasonable to have a single column per extension? this way `proj:code`, `proj:transform` could be stuffed into a single column and leave the possibility open for within-extension schema drift over time.

bitner: The main reason that we are promoting fields is to get the performance benefit of having them instantiated as postgres typed columns rather than as generic json, so there would be no point of just splitting up the json.

gadomski (commented):
Maybe I missed it, but can you add a table showing the "reference version" for each extension?

bitner:
> Maybe I missed it, but can you add a table showing the "reference version" for each extension?

@gadomski just added a doc at docs/src/promoted-fields.md

gadomski (approved):

PR #456, opened 2026-06-26
feat(search): v0.10 keyset search consolidation

bitner:
- Use an indexed query based on a worst case envelope calculated from the cql2 filter (max datetime, end_datetime, collections, and spatial extent envelope possible from the cql2) against partition statistics to get a histogram and min/max datetime from partitions that could be covered and a list of possible collections. These are then used to add filters to the query that don't change the filtering at all (they are guaranteed to be wider than the actual filters) which serve to help prune partitions making planning and execution faster. This replaces (and is faster than) the difficult to maintain check constraints that were added to partitions just for the partition pruning. The histogram also allows for smarter batch sizes when we chunk through datetime sorted results.
-  Add streaming data helpers search_plan() and collection_search_plan() to assist in building a streaming rust side client that can do the heavy lifting on the client side and drastically reduces the memory footprint for larger queries.
- Many optimizations to the hydration process including smarter handling of fields that we can skip due to include/exclude definitions.
- Return json rather than jsonb from item collection returning endpoints. As we don't need the functionality of jsonb at that stage, assembling json is faster and has a maximum size limit of 1GB rather than 256MB of jsonb.
- Integrate new chunking throughout all search endpoints including geometrysearch.
- Bring collection_search more inline with search. Add keyset pagination rather than limit offset.
- Change tokens used for keyset pagination to an encoded list of the values that are necessary to build the query for the next page. Previously we passed the collection and id which we then needed to create a separate query to get the values to create the filter.


src/pgstac/sql/000_idempotent_pre.sql
```diff
@@ -96,10 +96,8 @@ $$ LANGUAGE PLPGSQL;
 SELECT pgstac_admin_owns();
 
 CREATE SCHEMA IF NOT EXISTS pgstac AUTHORIZATION pgstac_admin;
-
-GRANT ALL ON ALL FUNCTIONS IN SCHEMA pgstac to pgstac_admin;
-GRANT ALL ON ALL TABLES IN SCHEMA pgstac to pgstac_admin;
-GRANT ALL ON ALL SEQUENCES IN SCHEMA pgstac to pgstac_admin;
+-- pgstac_admin owns the schema and all objects in it (pgstac_admin_owns() above +
+-- AUTHORIZATION), so it already has every privilege — no explicit GRANT ALL needed.
```
gadomski, line 100: This reads a little like AI-explainer slop; there's no GRANT ALL in the code anymore, so mentioning it feels weird?

src/pgstac/sql/002b_cql.sql
```diff
@@ -144,6 +144,98 @@ BEGIN
...
+-- q_op_query: SQL predicate for the pgstac `q` full-text operator. args is the search term(s)
+-- (a string or array of strings, the STAC `q` parameter); it is matched against a tsvector
+-- built from the row's description/title/keywords. Modeled on spatial_op_query /
+-- temporal_op_query so full-text is a first-class CQL2 op.
+CREATE OR REPLACE FUNCTION q_op_query(args jsonb) RETURNS text AS $$
+    SELECT format(
+        $q$(
+            to_tsvector('english', coalesce(properties->>'description', '')) ||
+            to_tsvector('english', coalesce(properties->>'title', '')) ||
+            to_tsvector('english', coalesce(properties->>'keywords', ''))
```
gadomski, line 233: Curiosity question, how hard would it be to allow folks to deploy versions of **pgstac** against other languages (i.e. make this and any other "english" references configurable)?

src/pgstac/sql/002b_cql.sql
```diff
@@ -198,6 +290,11 @@ BEGIN
     END IF;
 
     IF jsonb_typeof(j) = 'object' THEN
+        -- GeoJSON geometry args (Point/Polygon/.../GeometryCollection) are not cql1 expressions;
```
gadomski: `cql1`?

```suggestion
        -- GeoJSON geometry args (Point/Polygon/.../GeometryCollection) are not cql2 expressions;
```

gadomski (approved):
I'll be honest that I don't grok all of the search changes, but what I could understand makes sense. @bitner you mention performance gains from the enveloping, could/should we add some documentation of the benchmarks, to point to in the release notes?

PR #458, opened 2026-06-30
pgstac-rs: Rust engine for pgstac (search, ingest, dump/export)

bitner:
### Major reworking of io functions and tooling for pgstac.

- Add rustac based rust library with python bindings and cli for all public pgstac functionality
- SQL functions and rust tooling to enable streaming data from searches which improves performance and massively reduces memory for search returns.
- Remove the additional constraints that were used to enhance partition pruning. Replaced with an indexed lookup to a statistics table. This reduces lock contention and adds the ability to further prune partitions by spatial extent as well as maintaining the performance previously gained from the constraint on end_datetime.
- Numerous changes to increase robustness under heavy concurrency both on the SQL side and the Rust side.


src/pgstac-rs/Cargo.toml
```diff
@@ -11,21 +11,107 @@ repository = "https://github.com/stac-utils/pgstac"
...
+[[bin]]
+name = "pgstac"
+path = "src/bin/pgstac.rs"
+required-features = ["cli"]
+
+[[bin]]
+name = "memprofile"
+path = "src/bin/memprofile.rs"
+required-features = ["profiling"]
+
```
gadomski: Shipping to different binaries feels a little weird. It'd feel more natural to either:

- Put the memprofile behavior in the main CLI, or
- Make memprofile an [example](https://doc.rust-lang.org/cargo/guide/project-layout.html) to be able to build and use it in the repo, but not have to worry about other folks installing it.

src/pgstac-rs/src/lib.rs
```diff
@@ -218,14 +307,39 @@ pub trait Pgstac: GenericClient {
...
+    async fn collections(&self) -> Result<Vec<JsonValue>>
+    where
+        Self: Sized,
+    {
+        let mut all = Vec::new();
+        let mut token: Option<String> = None;
+        loop {
+            let page =
+                collections::collection_search(self, &serde_json::json!({}), token.as_deref())
+                    .await?;
```
gadomski: Passing this `self` to a free function implies (to me) that the function should be a method on `Pgstac` (like it was before)?

bitner:
## Current benchmarking run on local laptop

- TTLR = Time to Last Record  
- TTFR = Time to First Record (really only applicable to the ndjson streaming output from rust)  
- `911h` = v0.9.11 running `sql search()`  
- `911nh` = v0.9.11 running `sql search()` with `nohydrate`  
- `010sql` = v0.10.0 running `sql search()`  
- `010ic` = v0.10.0 using rust streaming behind the scenes to build a FeatureCollection  
- `ttfr/ttlr` = v0.10.0 using rust streaming, time to first record and time to last record  
- `010gpq` = v0.10.0 streaming arrow batches to stac-geoparquet  

Note: in all instances that return a full FeatureCollection, the entire set must be held in memory. Rust streaming in v0.10.0 stays very low constant memory. Rust streaming to geoparquet stays constant memory, but memory usage could still be high depending on row group size settings. Also, note that 911nh is still unhydrated, so any use of that path would still require the additional client side hydration, so it is not entirely apples to apples.

---

### pgstac search benchmark — bench_v0911 (0.9.11) vs bench_current (0.10), 100k items/collection  
_All values are TTLR ms (lower = faster); `rust` column is NDJSON-stream `TTFR/TTLR`; `010gpq` writes parquet._

| filter | limit | 911h | 911nh | 010sql | 010ic | rust TTFR/TTLR | 010gpq |
|---|---|---|---|---|---|---|---|
| broad | 10 | 190 | 58 | 92 | 87 | 53/55 | 60 |
| bbox CONUS | 10 | 199 | 82 | 126 | 55 | 52/54 | 70 |
| intersects poly | 10 | 184 | 57 | 107 | 55 | 43/49 | 73 |
| landsat 1-month | 10 | 165 | 57 | 83 | 51 | 62/64 | 75 |
| cql2 cloud<20 | 10 | 174 | 70 | 91 | 64 | 58/60 | 81 |
| sort cloud asc | 10 | 982 | 650 | 182 | 95 | 87/91 | 87 |
| fields id,geom | 10 | 62 | 58 | 57 | 47 | 44/45 | 48 |
| fields -assets | 10 | 69 | 74 | 68 | 53 | 53/55 | 55 |
| broad | 100 | 1097 | 103 | 195 | 122 | 44/59 | 125 |
| bbox CONUS | 100 | 1275 | 176 | 252 | 111 | 68/83 | 160 |
| intersects poly | 100 | 1082 | 148 | 237 | 129 | 56/91 | 185 |
| landsat 1-month | 100 | 1113 | 93 | 197 | 113 | 51/62 | 137 |
| cql2 cloud<20 | 100 | 842 | 89 | 167 | 104 | 50/61 | 100 |
| sort cloud asc | 100 | 1338 | 658 | 212 | 134 | 83/97 | 146 |
| fields id,geom | 100 | 99 | 92 | 105 | 44 | 51/56 | 56 |
| fields -assets | 100 | 138 | 128 | 264 | 61 | 61/74 | 86 |
| broad | 1000 | 10768 | 560 | 1290 | 654 | 50/140 | 841 |
| broad | 10000 | 102649 | 5914 | 11841 | 6040 | 66/984 | 6029 |

---

### PAGING — total ms to fetch K pages of L items via keyset tokens, vs one streaming pass of K*L

| filter | K x L | 911 pages | 010 pages | 010ic pages | stream K*L |
|---|---|---|---|---|---|
| broad | 5 x 100 | 5363 | 939 | 637 | 102 |
| bbox CONUS | 5 x 100 | 5514 | 967 | 615 | 116 |
| cql2 cloud<20 | 5 x 100 | 4195 | 843 | 549 | 95 |

---

### INGEST throughput — v0.9.11 pypgstac load vs the 0.10 Rust loader (from ingest_bench.sh, 100k/coll)

| collection | v0.9.11 ndjson | Rust ndjson | speedup | Rust parquet |
|---|---|---|---|---|
| naip 89.8k | 8,394 it/s | 26,758 it/s | 3.2x | 20,310 it/s |
| landsat 100k | 1,695 it/s | 4,231 it/s | 2.5x | 2,800 it/s |
| sentinel 100k | 2,710 it/s | 5,948 it/s | 2.2x | 4,191 it/s |

docs/src/pgstac-rs.md
```diff
@@ -0,0 +1,160 @@
...
+
+Pool methods: `search_page`, `search_items` (stream), `search_collect_items`, `stream_ndjson`,
+`search_matched`, `collection_search`, `get_item`, `get_collection`, `get_queryables`.
+
+## CLI
+
+```text
+pgstac <COMMAND>
+  dump      Dump a pgstac instance (or a subset) to a directory, tar, S3, or stdout
+  search    Stream a search/CQL2 result as NDJSON / ItemCollection / geoparquet (0.10 only)
```
gadomski, line 93: Will this be redundant with `rustac search`?

src/pgstac-rs/src/export/mod.rs
```diff
@@ -0,0 +1,124 @@
+//! Data export (dump) library for pgstac.
```
gadomski:
```suggestion
//! Data export (dump) module for pgstac.
```

src/pgstac-rs/src/read/search.rs
```diff
@@ -0,0 +1,451 @@
...
+
+use crate::Result;
+use crate::hydrate::{HydrationModel, Hydrator};
+use crate::keyset;
+use crate::source::{
+    PromotedSchema, detect_hydration_model, fetch_fragments, load_collection_context,
+    read_base_item_row, read_fragment_row,
+};
+use chrono::{DateTime, TimeZone, Utc};
+use serde_json::{Value, json};
```
gadomski, line 24: I'm surprised we don't use `stac::api::Search`? https://docs.rs/stac/latest/stac/api/struct.Search.html

gadomski (commented):
I didn't do a detailed review yet, just looked at some basic things. I've got a high level question about whether there's more stuff we can use from the `stac` crate (e.g. `Search`).

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1201 @@
...
+}
+
+/// Default ingest parallelism: CPU count capped at 8 (a safe ceiling for decode/connection fan-out).
+/// Powers both `--concurrency` (parallel load batches) and the ndjson decode-thread default, so the loader
+/// parallelizes out of the box — profiling showed single-threaded decode was the big-item ingest floor.
+fn default_ingest_parallelism() -> usize {
+    std::thread::available_parallelism()
+        .map(|n| n.get())
+        .unwrap_or(4)
+        .min(8)
```
gadomski: Should be module-level constants, just to make it clearer where they're set.

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1201 @@
...
+    #[arg(long)]
+    dry_run: bool,
+}
+
+#[derive(ValueEnum, Clone, Copy, Debug)]
+enum CompressionArg {
+    Zstd,
+    Snappy,
+    Uncompressed,
+}
```
gadomski: This has an enum in `rustac`, can we use that?

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1201 @@
...
+
+#[derive(ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
+enum SearchFormat {
+    /// Newline-delimited JSON, one item per line (streams all pages).
+    Ndjson,
+    /// A single STAC FeatureCollection page (SQL-faithful next/prev + context).
+    Itemcollection,
+    /// stac-geoparquet of the (bounded) result.
+    Geoparquet,
+}
```
gadomski: Same as below, can we re-use the enum from `rustac`?

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1201 @@
...
+
+impl From<LoadPolicy> for ConflictPolicy {
+    fn from(p: LoadPolicy) -> Self {
+        match p {
+            LoadPolicy::Upsert => ConflictPolicy::Upsert,
+            LoadPolicy::Ignore => ConflictPolicy::Ignore,
+            LoadPolicy::Error => ConflictPolicy::Error,
+        }
+    }
+}
```
gadomski: Is this really only CLI-specific? Also, is it actually different from `ConflictPolicy`? Can we re-use?

src/pgstac-rs/src/load/ingest.rs
```diff
@@ -0,0 +1,956 @@
...
+/// [`crate::PgstacPool::create_items`]. A cheap pass (id + datetime, plus a content hash for upsert/delsert)
+/// buckets items by partition window; [`precheck_one_partition`] classifies each partition's bucket in
+/// parallel, and the survivors take the normal load (resolving same-partition conflicts per `policy`).
+/// `ignore` skips every existing id; `upsert`/`delsert` skip only ids whose content is unchanged.
+///
+/// Matching is by id: a datetime change within a partition is caught, but a cross-partition move reads as
+/// new (the old row lives in another partition) — `Upsert` leaves that old row in place, `Delsert` removes
+/// it. Returns `(unchanged_skipped, loaded)`.
+pub async fn precheck_upsert(
+    pool: &crate::PgstacPool,
```
gadomski, line 541: This is gated behind the "pool" feature, so either needs a `#[cfg(feature = "pool")]` or some other fix.

We should add CI that runs with and without the `pool` feature to catch stuff like this.

src/pgstac-rs/src/read/feature.rs
```diff
@@ -0,0 +1,266 @@
+//! Byte-assembly output for hydrated 0.10 items.
+//!
+//! [`write_fragment_feature`] writes a fully-hydrated STAC item straight to a writer without ever
+//! materializing the merged `assets` as a [`Value`]: the item's `assets` are read as raw JSON, the
+//! shared fragment's `assets` are borrowed (by `&`, never cloned per item), and the two are merged
+//! **at serialize time** — descending only where both sides are objects at a key, and emitting arrays,
+//! scalars, and one-sided keys **verbatim** from the raw bytes. `bbox` is likewise emitted from raw
+//! bytes (preserving PostgreSQL `numeric` precision that an f64 round-trip would lose). Everything else
+//! reuses [`hydrate_fragment_core`].
```
gadomski: This is a private function, so the docs won't link properly.

src/pgstac-rs/src/db/client.rs
```diff
@@ -0,0 +1,227 @@
...
+impl PgConn for deadpool_postgres::Client {
+    fn pg(&self) -> &tokio_postgres::Client {
+        self
+    }
+    fn pg_mut(&mut self) -> &mut tokio_postgres::Client {
+        self
+    }
+}
+
+/// A wrapper around a [`GenericClient`] that implements the STAC client traits ([`ItemsClient`],
```
gadomski: There's no such thing as a `GenericClient`?

src/pgstac-rs/src/load/dehydrate.rs
```diff
@@ -0,0 +1,409 @@
...
+    /// `items.link_hrefs` — the per-link `href`s (NULL entries preserved).
+    pub link_hrefs: Option<Vec<Option<String>>>,
+    /// Promoted queryable columns, in schema order, aligned with `schema.columns()`.
+    pub promoted: Vec<PromotedValue>,
+}
+
+/// Dehydrates one self-contained STAC item into a [`DehydratedRow`], consuming `item`.
+///
+/// Matches SQL `content_dehydrate(item)`: computes the canonical `item_hash` over the whole item, splits
+/// the geometry to EWKB, derives the temporal columns ([`stac_daterange`] semantics), promotes the
```
gadomski: Unresolved doc link for `stac_daterange`

src/pgstac-rs/src/load/ingest.rs
```diff
@@ -0,0 +1,956 @@
...
+    ords: Vec<i32>,
+    ids: Vec<String>,
+    hashes: Vec<Vec<u8>>,
+}
+
+static PRECHECK_TEMP_SEQ: AtomicU64 = AtomicU64::new(0);
+
+/// Loads `items`, loading only those not already present-and-current and skipping the rest, via
+/// [`crate::PgstacPool::create_items`]. A cheap pass (id + datetime, plus a content hash for upsert/delsert)
+/// buckets items by partition window; [`precheck_one_partition`] classifies each partition's bucket in
```
gadomski: `precheck_one_partition` is a private function, so the docs will break.

src/pgstac-rs/src/lib.rs
gadomski: Need to update the features list to include `pool`

src/pgstac-rs/src/json/canonical.rs
```diff
@@ -0,0 +1,263 @@
...
+/// ```
+/// use serde_json::json;
+/// assert_eq!(pgstac::canonical::jsonb_canonical(&json!({"b": 1, "a": 2})), r#"{"a":2,"b":1}"#);
+/// ```
+pub fn jsonb_canonical(value: &Value) -> String {
+    let mut buf = String::new();
+    write_canonical(&mut buf, value);
+    buf
+}
+
```
gadomski: This function doesn't appear to be used anywhere? `jsonb_hash` is used in `dehydrate`, but this one isn't.

src/pgstac-rs/src/read/collections.rs
```diff
@@ -0,0 +1,128 @@
...
+///
+/// Mirrors the SQL `collection_search` paging contract (including the `prev` reversal) but does the
+/// work client-side off `collection_search_plan`. The returned [`SearchPage`] holds the collections in
+/// `features`.
+pub async fn collection_search<C: GenericClient>(
+    client: &C,
+    search: &Value,
+    token: Option<&str>,
+) -> Result<SearchPage> {
+    let is_prev = token.is_some_and(|t| t.starts_with("prev:"));
```
gadomski, line 23: I still think these could be methods provided by the `Pgstac` trait, since they take a client as the first argument?

src/pgstac-rs/src/read/collections.rs
```diff
@@ -0,0 +1,128 @@
...
+        prev_token,
+        number_matched,
+    })
+}
+
+/// Fetches a single collection by id (wraps [`collection_search`] with an id filter).
+pub async fn get_collection<C: GenericClient>(
+    client: &C,
+    collection_id: &str,
+) -> Result<Option<Value>> {
```
gadomski, line 117: Any reason not to return a `stac::Collection` here?

src/pgstac-rs/src/load/dehydrate.rs
```diff
@@ -0,0 +1,409 @@
...
+        |what: &str| Error::Dehydrate(format!("item {id}: property {stac_name} is not {what}"));
+    Ok(match (kind, raw) {
+        (PromotedKind::Text, Some(v)) => PromotedValue::Text(Some(value_to_text(v))),
+        (PromotedKind::Text, None) => PromotedValue::Text(None),
+        (PromotedKind::Float, Some(v)) => {
+            PromotedValue::Float(Some(value_to_f64(&v).ok_or_else(|| bad("a number"))?))
+        }
+        (PromotedKind::Float, None) => PromotedValue::Float(None),
+        (PromotedKind::Int, Some(v)) => PromotedValue::Int(Some(
+            value_to_i64(&v).ok_or_else(|| bad("an integer"))? as i32,
```
gadomski: `as i32` can silently narrow ... I prefer erroring conversions (`try_into`).

src/pgstac-rs/src/export/planner.rs
```diff
@@ -0,0 +1,1218 @@
...
+            let queue = queue.clone();
+            let sink = sink.clone();
+            let budget = budget.clone();
+            let tx = tx.clone();
+            let snap_id = snapshot_id.clone();
+            let scan_filter = scan_filter.clone();
+            let factory_fut = factory.connect();
+            // Connect each worker up front so a connection failure is reported
+            // before the worker loop starts.
+            let client = factory_fut.await?;
```
gadomski, line 386: Claude flagged up that we don't clean up existing workers if this fails? Don't know if that matters in reality.

src/pgstac-rs/src/read/feature.rs
```diff
@@ -0,0 +1,266 @@
...
+    let (core, assets_raw, bbox_raw) = hydrate_fragment_core(item, fragment);
+    let feature = FeatureWriter {
+        core: &core,
+        bbox: bbox_raw.as_ref().map(|r| r.0.as_ref()),
+        frag_assets,
+        item_assets: assets_raw.as_ref().map(|r| r.0.as_ref()),
+    };
+    serde_json::to_writer(write, &feature)?;
+    Ok(())
+}
```
gadomski: From an API design perspective, this feels like it could be a method on `DehydratedItem`?

src/pgstac-rs/src/json/geom.rs
```diff
@@ -0,0 +1,117 @@
...
+    let geojson = serde_json::to_string(geometry)?;
+    Ok(GeoJson(&geojson).to_ewkb(CoordDimensions::xy(), Some(4326))?)
+}
+
+/// The raw EWKB bytes of an `items.geometry` value, read straight off the binary protocol.
+///
+/// Keeping the bytes lets the dump path forward WKB to Arrow/Parquet untouched, while the JSON path
+/// converts to GeoJSON via [`RawGeometry::to_geojson`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RawGeometry(pub Vec<u8>);
```
gadomski: Nit: Why not just name it `Ewkb`?

src/pgstac-rs/src/load/ingest.rs
```diff
@@ -0,0 +1,956 @@
...
+use std::error::Error as StdError;
+use std::sync::atomic::{AtomicU64, Ordering};
+use tokio_postgres::binary_copy::BinaryCopyInWriter;
+use tokio_postgres::types::{IsNull, ToSql, Type, to_sql_checked};
+
+/// EWKB bytes wrapped so they encode as a PostGIS `geometry` on a binary COPY.
+///
+/// The geometry binary wire format is EWKB, so the bytes are written verbatim.
+#[derive(Debug, Clone)]
+pub struct WkbGeometry(pub Vec<u8>);
```
gadomski: Duplicative on the geometry from the `geom` module?

src/pgstac-rs/src/read/keyset.rs
gadomski, line 1: The keyset stuff feels non-pgstac-specific, so a candidate for eventually pushing down to **rustac**?

src/pgstac-rs/src/db/client.rs
```diff
@@ -0,0 +1,227 @@
...
+/// });
+/// let client = Client::new(pg_client);
+/// let item_collection = client.search(Default::default()).await.unwrap();
+/// # })
+/// ```
+#[derive(Debug)]
+pub struct Client<C> {
+    client: C,
+    hydration: Arc<OnceCell<CachedHydration>>,
+}
```
gadomski, line 80: Could/should we get rid of the `Pgstac` trait, if this is the better way to do stuff? I got confused while reading the docstring for this struct, and if I got confused, other folks might as well.

src/pgstac-rs/src/db/connect.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub sslmode: Option<String>,
+    /// Path to a trusted CA certificate (`PGSSLROOTCERT`); consumed by the TLS connector.
+    pub sslrootcert: Option<String>,
+    /// Path to a client certificate (`PGSSLCERT`); consumed by the TLS connector.
+    pub sslcert: Option<String>,
+    /// Path to a client private key (`PGSSLKEY`); consumed by the TLS connector.
+    pub sslkey: Option<String>,
+    /// The search path applied at connection startup. Defaults to [`DEFAULT_SEARCH_PATH`].
+    pub search_path: String,
+}
```
gadomski, line 65: This feels _really_ duplicative on stuff that already exists for postgres, do we really need it?

src/pgstac-rs/src/read/page.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Links
+    ///
+    /// Added in pgstac v0.9
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub links: Vec<Link>,
+
+    /// Additional fields.
+    #[serde(flatten)]
+    pub additional_fields: Map<String, Value>,
+}
```
gadomski: Is this duplicative on `SearchPage`?

gadomski (changes requested):
This was a API-level review, meaning that I looked through the publicly exposed functions (via the documentation). I have not looked through the function bodies, except when I found myself digging deeper.

There's a lot of free functions that feel like they could/should be methods provided by the `Pgstac` trait or the `Client` struct, since they take a `client` as their first argument?

Also, Claude flagged up a couple of issues:

- https://github.com/stac-utils/pgstac/blob/4400bfd6b5e1796975bd612f9791547159791a2c/src/pypgstac/src/pypgstac/load.py#L480 is broken as `update_partition_stats_q` doesn't exist anymore
- Is it intentional that `partition_stats` isn't backfilled during a migration?
- Is `indexes_pending` used anywhere to build the queryable indexes?

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1216 @@
...
+    /// Delete an item (with --item) or a whole collection (without --item).
+    Delete(DeleteArgs),
+}
+
+#[derive(clap::Args, Debug)]
+struct DeleteArgs {
+    /// Postgres connection string. Defaults to $PGSTAC_DSN or a local dev DSN.
+    #[arg(long, env = "PGSTAC_DSN")]
+    dsn: Option<String>,
+
```
gadomski: You can use `global = true` in `arg` to share `dsn` among all the argument structures, e.g. https://github.com/stac-utils/rustac/blob/main/crates/cli/src/lib.rs#L43C1-L57C34

src/pgstac-rs/src/bin/pgstac.rs
```diff
@@ -0,0 +1,1216 @@
...
+    /// Collection id (required).
+    #[arg(long, short)]
+    collection: String,
+
+    /// Item id. Given, deletes that item; omitted, deletes the whole collection (and its items).
+    #[arg(long)]
+    item: Option<String>,
+}
```
gadomski, line 81: Since `delete` is a dangerous operation, should we add a confirmation prompt by default that could be disabled with a boolean flag?

src/pgstac-rs/src/export/budget.rs
```diff
@@ -0,0 +1,203 @@
...
+        return total;
+    }
+    FALLBACK
+}
+
+/// cgroup v2 `memory.max`, then v1 `memory.limit_in_bytes`. A "max" / very large
+/// value (unlimited) is treated as no cgroup limit.
+fn cgroup_memory_limit() -> Option<u64> {
+    // cgroup v2
+    if let Ok(s) = std::fs::read_to_string("/sys/fs/cgroup/memory.max") {
```
gadomski, line 127: This feels pretty gnarly — does it work on non-linux systems?

src/pgstac-rs/src/export/sink.rs
```diff
@@ -0,0 +1,461 @@
...
+/// type and the [`Sink`] trait offers a blocking bridge used by the planner.
+///
+/// The methods take a *relative path within the dump root* (e.g.
+/// `collections/landsat-c2-l2/202401.parquet`).
+pub trait Sink: Send + Sync {
+    /// Writes a whole small file from an in-memory buffer.
+    fn put(&self, rel_path: &str, bytes: &[u8]) -> Result<FileWritten>;
+
+    /// Writes a finished large file by streaming it from a local path.
+    fn put_file(&self, rel_path: &str, src: &Path) -> Result<FileWritten>;
```
gadomski:
```suggestion
    fn put_from_path(&self, rel_path: &str, src: &Path) -> Result<FileWritten>;
```

src/pgstac-rs/src/export/sink.rs
```diff
@@ -0,0 +1,461 @@
...
+            sha256: hasher.finalize_hex(),
+        })
+    }
+}
+
+// ---------------------------------------------------------------------------
+// object_store sink (local dir + S3/GCS/Azure)
+// ---------------------------------------------------------------------------
+
+#[cfg(feature = "cli")]
```
gadomski: Since this doesn't have to be just for the CLI, _to me_ this should be behind an `object_store` (or `store`) feature, and then the `cli` feature can depend on the `object_store` feature.

src/pgstac-rs/src/export/sink.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    /// Streams to any `object_store` backend (local fs, S3, GCS, Azure).
+    ///
+    /// Construct from a base URL (`file:///abs/dir`, `s3://bucket/prefix`, …);
+    /// relative dump paths are joined onto the base prefix. Large files use
+    /// multipart upload.
+    pub struct ObjectStoreSink {
+        store: Arc<dyn ObjectStore>,
+        base_prefix: ObjPath,
+        runtime: tokio::runtime::Handle,
```
gadomski: Why does this own a runtime? Can we just make the methods async on the trait?

gadomski (changes requested):
- Some comments from my previous PR review need to be addressed, e.g. around enum re-use and using some **stac** structures
- Got a local error running the test `canonical_and_hash_match_sql`: `function pgstac.jsonb_canonical_hash(jsonb) does not exist`

src/pgstac-rs/src/lib.rs
bitner: ok, this should be fixed

src/pgstac-rs/src/read/search.rs
bitner: we use stac::api::Search now

src/pgstac-rs/src/load/ingest.rs
bitner: done.

src/pgstac-rs/src/read/keyset.rs
bitner: I think that yes, in the future we might want to move this upstream and combine with the ability to take the keyset + a sort config and turn it into a CQL2 filter that we can apply to a next/prev page, but I think it's bespoke enough to pgstac right now to leave it here.

src/pgstac-rs/src/bin/pgstac.rs
bitner: We're now directly using compression string parsing from upstream. The only difference we have here is for the dump mode we have the default set to zstd-1. Defaults for search remain coming from upstream. Running some benchmarks there, the time difference really does make up for a not too big a difference in size.

src/pgstac-rs/src/bin/pgstac.rs
bitner: Everything should be coming from upstream now.

src/pgstac-rs/src/load/ingest.rs
bitner: Everything centralized on Ewkb now with from/to

src/pgstac-rs/src/db/connect.rs
bitner: Changed now so that only the things that differ (our custom search path mangling and some ssl stuff that isn't in the upstream postgres config) are in there.

src/pgstac-rs/src/bin/pgstac.rs
bitner: done.

src/pgstac-rs/src/export/budget.rs
bitner: ok, that is behind a feature now for linux and should gracefully degrade if it can't get the cgroup memory

src/pgstac-rs/src/export/sink.rs
bitner: changed

src/pgstac-rs/src/export/sink.rs
bitner: now behind a store feature

src/pgstac-rs/src/export/sink.rs
bitner: now async methods on the trait

gadomski:
@bitner can we get https://github.com/stac-utils/rustac/pull/1077 merged first? That way we don't have to undo a pin after release.

bitner:
> @bitner can we get https://github.com/stac-utils/rustac/pull/1077 merged first? That way we don't have to undo a pin after release.

we'll definitely need it merged before the actual .10 release, but pinned the branch now as the only way to get ci to pass

docs/src/pypgstac.md
```diff
@@ -1,70 +1,69 @@
...
-
-For direct migration operations (recommended for new integrations):
+# pyPgSTAC has been removed
+
+The legacy `pypgstac` Python package (migrations, bulk loading, queryables, and
+the pure-Python hydration helpers) has been removed from this repository. Every
+capability it provided now lives in one of its replacements:
+
+- **`pgstac-migrate`** — the migration runtime (planning + apply), published to
+  PyPI as [`pgstac-migrate`](https://pypi.org/project/pgstac-migrate/).
```
gadomski, line 8: Feels a little strange that we have `pypgstac-rs` and `pgstac-migrate` — feels like we should commit to the `py*` prefix or drop it.

src/pgstac-rs/src/export/format.rs
```diff
@@ -0,0 +1,484 @@
...
+                        Value::Bool(b) => b.to_string(),
+                        Value::Number(n) => n.to_string(),
+                        _ => continue,
+                    };
+                    *v = Value::String(s);
+                }
+            }
+        }
+    }
+}
```
gadomski, line 229: This is duplicative on https://github.com/stac-utils/rustac/blob/705857958f1ced81277e2e674c7ba7e20eb09066/crates/core/src/item.rs#L571-L606 but maybe it's specific enough to keep?

gadomski (approved):
I gave it a pretty detailed readthrough and it seems good. I'll be up-front that I didn't read every singe line 🙈 just too big.

